PR #46, opened 2019-04-08
Filter outlier barcodes-minimizers

gokcedilek:


gokcedilek:
The iterator change we talked about  (for a little bit of speed up) does not work when ```counts``` is a robin map. I could keep ```counts``` as an unordered map, because I guess we wouldn't want to keep both the ```find``` method and the `[]` look up.

sjackman:
What's the compiler error message that you get from the robin map? Copy and paste it here, and surround copied-and-pasted code with triple back ticks. See [GitHub Help / Quoting Code](https://help.github.com/articles/basic-writing-and-formatting-syntax/#quoting-code).

gokcedilek:
I added (finally) filter-minimizers, which works exactly in the same workflow as the python one. (and I added a test to the Makefile, I can add more tests as well). No improvements or combining code yet; because we said we should first merge the identical version and then move onto the alternative version. I will also post a timing result for my cpp filter-barcodes + python filter-minimizers vs cpp filter-barcodes-minimizers, however I will probably do it tomorrow because I will first try to figure out how to use the numbers cluster and I will also re-do my benchmarkings from yesterday.

sjackman:
One style issue to sort out:
```
physlr-bxmx.cc:187:24: error: narrowing conversion from 'float' to 'std::vector::size_type' (aka 'unsigned long') [bugprone-narrowing-conversions,-warnings-as-errors]
                        qs.push_back(values[std::nearbyint(p*(values.size()-1))]);
```
Explicitly cast to `(size_t)` when casting from a real to an integer.

gokcedilek:
Is there a way to temporarily skip the ```run clang-tidy``` step? Just to see what the remaining will do.
When my code is:
```
auto mxit = counts.find(mx);
if (mxit == counts.end()) {
 mxit->second = 1;
} else {
 ++mxit->second;
} 
``` 
```the error is: error: assignment of member 'std::pair<long unsigned int, unsigned int>::second' in read-only object```

**Comment by gokcedilek** [comment]
And when my code is:
```
auto mxit = counts.find(mx);
if (mxit == counts.end()) {
 *mxit = 1;
 } else {
  ++*mxit;
}
```
the error is:
```
error: no match for 'operator++' (operand type is 'tsl::detail_robin_hash::robin_hash<std::pair<long unsigned int, unsigned int>, tsl::robin_map<long unsigned int, unsigned int>::KeySelect, tsl::robin_map<long unsigned int, unsigned int>::ValueSelect, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int, unsigned int> >, false, tsl::rh::power_of_two_growth_policy<2ul> >::robin_iterator<false>::value_type {aka const std::pair<long unsigned int, unsigned int>}')
     ++*mxit;
```

**Comment by gokcedilek** [comment]
aah okay I just understood what the error is -- I should write ```values[size_t(std::nearbyint(p*(values.size()-1)))``` right?

**Comment by sjackman** [comment]
Yep!

**Comment by gokcedilek** [comment]
I couldn't figure out why but this declaration doesn't work with the robin map but it works with the unordered map:
```using MxtoCount = tsl::robin_map<Mx, std::atomic<unsigned>>;```
This leads to an error in the declaration of the "counts" map which says ``` use of deleted function 'std::atomic<unsigned int>::atomic(const std::atomic<unsigned int>&)```
I could use the unordered_map for counts, or I could instead use a mutex lock. What do you think @sjackman @JustinChu ?

**Comment by JustinChu** [comment]
There is some context missing here what does the actual instantiation of the object look like?
Similar to this issue:
https://stackoverflow.com/questions/27314485/use-of-deleted-function-error-with-stdatomic-int

**Comment by gokcedilek** [comment]
This version is identical: I had a bug in calculating quantiles because I wasn't updating counts for the second time after removing singletons, plus I also had a bug in my print statements that I realised later on because I was incrementing singletons or frequent minimizers many times for the same minimizer, instead I just put them in a set and outputted the size of the set in this change. With these changes, I will re-do my Numbers timings and post in here! I also added two tests to the Makefile, one is with using the -C option (without quantiles) and one is without using the option (with quantiles).

**Comment by gokcedilek** [comment]
|  f1 | Time (min) | Memory (GB)  |  
| --- | --- | --- |
|  py-fb-py-fm | 110  |  190.155 |
|  cpp-fb-py-fm | 44  | 159.605  | 
|  cpp-fb-cpp-fm |  37 |  146.516 |

File: src/Makefile [hunk 1]
```diff
@@ -1,13 +1,10 @@
-CXXFLAGS=-g -O3 -Wall -Werror -Wno-error=unknown-pragmas -std=c++11 -pthread -I./include -I./vendor -fopenmp
-LDLIBS=-lz
+CXXFLAGS=-g -O2 -Wall -Werror -Wno-error=unknown-pragmas -std=c++11 -pthread -I./include -I./vendor -fopenmp
+LDLIBS=-lz
```
sjackman: Please revert the changes to this chunk.

src/data/checkMx.sh
```diff
@@ -16,13 +17,13 @@ fi
...
 	if [[ "$diff" != "" ]]; then
 		echo "${bx} has different minimizers!"
 		exit 1
 	else
 		rm ${dir}/${bx}.${cpp} ${dir}/${bx}.${py}
 	fi
 done < ${dir}/barcodes.$cpp
-
 rm ${dir}/barcodes.$cpp ${dir}/barcodes.$py
+touch "${py}_done"
```
sjackman: Is this line needed?

gokcedilek:
I made the changes, and I also did a small change so now it's just a little bit faster :). The change I did is after removing singletons, I was re-counting all the minimizers to update ```counts```, but now I'm just removing if the count is less than 2.

src/Makefile
sjackman: Please remove the stray white space here.
```suggestion
LDLIBS=-lz
```

sjackman:
Please squash and merge. Thank you for this work, and well done! 🌟

gokcedilek:
|  hg004 | Time (min) | Memory (kb)  |  
| --- | --- | --- |
|  py-fb-py-fm | 401  |  422804472 |
|  cpp-fb-fm | 90  | 295633064  |

sjackman:
4.5 fold improvement! Nice! 🏇

PR #48, opened 2019-04-25
My benchmarking work

gokcedilek:


sjackman:
Please remove the igraph commit 5a63ae65f0318538639430635f1b43dc811baf95 from this PR.

gokcedilek:
I just saw that and did that:) 
For the R script, we can add an if conditional to check if both the time and memory files were provided or not, right now it is plotting both of them.

gokcedilek:
Also, I have separate scripts to get the times, to convert from h:mm:ss format to minutes, and to convert to the tsv format with "begin interval end". To save time to do other stuff, I uploaded them as they are, however I/someone could combine them.

sjackman:
> For the R script, we can add an if conditional to check if both the time and memory files were provided or not, right now it is plotting both of them.

No need. We should always have both time and memory data in the `.time` files.

sjackman:
> Also, I have separate scripts to get the times, to convert from h:mm:ss format to minutes

See the `hms` function in https://lubridate.tidyverse.org

gokcedilek:
I will update this PR with:
1) Bash script to extract time and memory usages for each step into 2 files
2) R script to do everything else

gokcedilek:
hey Shaun :smile:, I pushed something from my personal computer; I realized something!!! If we use the ```read_tsv``` function, rather than the ```read.csv```  function with tab separator, the input file is read differently!!! In fact, it is read in a way that we don't want!
Example:
if the timing of the indexlr stage is 1:26
--> ```read.csv``` reads this as 1:26
--> ```read_tsv``` reads this is as 01:26:00 (I wonder why) therefore, the subsequent calculations go wrong.
Then for ggplot I converted the duration objects to numeric. I hope this should work now. I modified ```benchmarking.R```. 
This PR also has all my previous Bash scripts for this work, I can delete all of them except the one parsing the time and memory usages from given file names, or should I keep them?

sjackman:
> hey Shaun 😄, I pushed something from my personal computer; I realized something!!! If we use the read_tsv function, rather than the read.csv function with tab separator, the input file is read differently!!! In fact, it is read in a way that we don't want!

Interesting! Good find. `read_tsv` attempts to detect the type of each column. It's evidently detecting the type of that column as `hh:mm`, which is wrong. When it's guess is wrong, you can explicitly override it using the `col_types` parameter of `read_tsv`. For example `col_types = "cc"` to specify two columns, each of which are character columns (instead of a time column). Then we can do the conversion ourself from character to time.

See https://readr.tidyverse.org/reference/read_delim.html

sjackman:
> This PR also has all my previous Bash scripts for this work, I can delete all of them except the one parsing the time and memory usages from given file names, or should I keep them?

You can delete them if they're no longer needed.

gokcedilek:
very interesting, reading as a character changes the default interpretation! I removed the bash scripts except the one to write time and memory as tsv files:)

sjackman:
I don't think you intended to delete `src/README.md`.
Could you please both the R script and shell script to the `data/bin/` directory?

gokcedilek:
I'll try to find the commit that I deleted ```src/README.md``` and revert it back.

sjackman:
Looks like the file `src/README.md` was deleted in commit 820fbca2e6d8345c625b1a5425d8f0322ceda322. Try…
```sh
git checkout timing
git fetch origin
git rebase -i origin/master
# delete the line `pick 820fbca benchmarks script`
# save and quit
git push origin timing --force
```

gokcedilek:
woww deleting from rebasing is a great method! it worked, thank you for the suggestion. to find which commit the change belongs to (not from Github, but from the command line), is there a better way than doing ```git log --pretty --oneline```? also, in order to keep a file in local repo but remove from remote repo (because I was curious about doing this), this triplet worked after some trial-error:
```
git rm --cached <file_name>
git commit --amend
git push origin master --force
```

sjackman:
> to find which commit the change belongs to (not from Github, but from the command line), is there a better way than doing git log --pretty --oneline?

```sh
git log --pretty --oneline -- src/README.md
```

> also, in order to keep a file in local repo but remove from remote repo (because I was curious about doing this), this triplet worked after some trial-error:

Nice! I hadn't used that command before.

PR #50, opened 2019-04-26
reformatting of the code

afshinfard:
Just did some small rearrangements of the code


physlr/physlr.py
```diff
@@ -1149,9 +1149,9 @@ def physlr_molecules(self):
             strategy_switcher.get(self.args.strategy,
                                   "\033[93m"+"\n\tWarning:"
```
```suggestion
                                  "\033[93m\n\tWarning:"
```

afshinfard:
Thanks @sjackman, you too!

PR #52, opened 2019-04-27
Determine_molecules sub methods modularization

afshinfard:


physlr/physlr.py
```diff
@@ -1021,107 +1020,127 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
-        if strategy == 3:
-            return Physlr.determine_molecules_louvain(g, u)
-        if strategy == 4:
-            return Physlr.determine_molecules_cosine_of_squared(g, u)
+        """Assign the neighbours of this vertex to molecules."""
+        communities = []
+        if strategy not in {2, 3, 4, 5, 10}:
+            # default strategy (bi-connected components) in case of wrong parameter.
+            communities = \
+                Physlr.community_detection_biconnected_components(g, set(g.neighbors(u)))
```
sjackman: This should produce an error message if the parameter value is not valid.

physlr/physlr.py
```diff
@@ -1021,107 +1020,127 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
 
-        # strategy == 1 or none of the previous strategies
-        return Physlr.determine_molecules_biconnected_components(g, u)
+        if strategy == 2:  # bi-connected + k-clique
+            communities = Physlr.determine_molecules_bc_k_cliques(g, u)
+
```
sjackman: You can remove these blank lines.

physlr/physlr.py
```diff
@@ -1021,107 +1020,127 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
+        if strategy == 3:  # bi-connected + Louvain
```
```suggestion
        elif strategy == 3:  # bi-connected + Louvain
```

physlr/physlr.py
```diff
@@ -1021,107 +1020,127 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
+            communities = Physlr.determine_molecules_bc_louvain(g, u)
+
+        if strategy == 4:  # bi-connected + sqCos
```
```suggestion
        elif strategy == 4:  # bi-connected + sqCos
```

physlr/physlr.py
```diff
@@ -1021,107 +1020,127 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
+            communities = Physlr.determine_molecules_bc_cosine_of_squared(g, u)
+
```
```suggestion
        else:
          # print an error message and exit failure
```

physlr/physlr.py
```diff
@@ -1021,107 +1020,127 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
+                partition = louvain.best_partition(g.subgraph(node_set))
+            else:
+                partition = louvain.best_partition(g.subgraph(node_set), init_communities)
+            #     communities = []
+            #     for com in set(partition.values()):
+            #         community_nodes = \
+            #             {nodes for nodes in partition.keys() if partition[nodes] == com}
+            #         if len(community_nodes) > 1:
+            #             communities.append(community_nodes)
+            # return communities
```
sjackman: Don't commit commented-out code. Delete it if it's not being used.

physlr/physlr.py
```diff
@@ -1021,107 +1020,127 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
         return "_" + str(random.choice(max_hits)), 0, 1
 
     @staticmethod
-    def determine_molecules_biconnected_components(g, u):
-        "Separate bi-connected components."
-        cut_vertices = set(nx.articulation_points(g.subgraph(g.neighbors(u))))
-        components = list(nx.connected_components(g.subgraph(set(g.neighbors(u)) - cut_vertices)))
+    def community_detection_biconnected_components(g, node_set):
```
```suggestion
    def detect_communities_biconnected_components(g, node_set):
```
Methods should be named `verb_noun`, so `detect_communities_biconnected_components`.

physlr/physlr.py
afshinfard: aha, and what function should I use to halt/exit the program @sjackman ? Should I call it inside this function?

sjackman: Yes. `exit(1)`

afshinfard:
Changes are made @sjackman, I appreciate if you review.

physlr/physlr.py
```diff
@@ -1021,107 +1020,118 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
-        communities = []
-        for comp in components:
-            if len(comp) > 1:
-                partition = louvain.best_partition(g.subgraph(comp))
-                for com in set(partition.values()):
-                    list_nodes = [nodes for nodes in partition.keys() if partition[nodes] == com]
-                    if len(list_nodes) > 1:
-                        communities.append(list_nodes)
-        return u, {v: i for i, vs in enumerate(communities) if len(vs) > 1 for v in vs}
+        if len(node_set) > 1:
```
```suggestion
```
I don't think this test is needed.

physlr/physlr.py
```diff
@@ -1021,107 +1020,118 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
+            return [set(i) for i in nxcommunity.k_clique_communities(g.subgraph(node_set), k)]
```
```suggestion
          return list(nx.community.k_clique_communities(g.subgraph(node_set), k))
```

physlr/physlr.py
```diff
@@ -1021,107 +1020,118 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
+        return []
```
```suggestion
```

physlr/physlr.py
```diff
@@ -1021,107 +1020,118 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
-        if not partition:
-            return u, {}
-        multinode_partitions_set = {}
-        for com in set(partition.values()):
-            list_nodes = [nodes for nodes in partition.keys() if partition[nodes] == com]
-            if len(list_nodes) > 1:
-                multinode_partition = {v: com for v in list_nodes}
-                multinode_partitions_set.update(multinode_partition)
-        return u, multinode_partitions_set
+        if len(node_set) > 1:
```
sjackman: Please remove these special cases wherever they are not truly needed.

physlr/physlr.py
```diff
@@ -1021,107 +1020,118 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
-                        communities.append(list_nodes)
-        return u, {v: i for i, vs in enumerate(communities) if len(vs) > 1 for v in vs}
+        if len(node_set) > 1:
+            return [set(i) for i in nxcommunity.k_clique_communities(g.subgraph(node_set), k)]
+        return []
 
     @staticmethod
-    def determine_molecules_just_louvain(g, u):
-        "Apply louvain community detection without bi-connected separation."
+    def detect_communities_louvain(g, node_set, init_communities=False):
```
```suggestion
    def detect_communities_louvain(g, node_set, init_communities=None):
```

physlr/physlr.py
```diff
@@ -1021,107 +1020,118 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
-            list_nodes = [nodes for nodes in partition.keys() if partition[nodes] == com]
-            if len(list_nodes) > 1:
-                multinode_partition = {v: com for v in list_nodes}
-                multinode_partitions_set.update(multinode_partition)
-        return u, multinode_partitions_set
+        if len(node_set) > 1:
+            if not init_communities:
+                partition = louvain.best_partition(g.subgraph(node_set))
+            else:
+                partition = louvain.best_partition(g.subgraph(node_set), init_communities)
```
```suggestion
              partition = louvain.best_partition(g.subgraph(node_set), init_communities)
```
Replace these four lines with this one line.

physlr/physlr.py
```diff
@@ -1021,107 +1020,118 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
+            return [{nodes for nodes in partition.keys() if partition[nodes] == com}
```
```suggestion
            return [{u for u in partition.keys() if partition[u] == com}
```
`nodes` is I believe a single node, yes? If so, it the variable name should be `node` or `u` rather than `nodes`.

physlr/physlr.py
```diff
@@ -1021,107 +1020,118 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
-                                           for i, j in edges_to_remove]
-                sub_graph_copy = nx.Graph(sub_graph)
-                sub_graph_copy.remove_edges_from(edges_to_remove_barcode)
-                cos_components = list(nx.connected_components(sub_graph_copy))
-                cos_components.sort(key=len, reverse=True)
-                for com in cos_components:
-                    if len(com) > 1:
-                        communities.append(com)
-        return u, {v: i for i, vs in enumerate(communities) if len(vs) > 1 for v in vs}
+        if len(node_set) > 1:
```
sjackman: Remove special cases unless truly needed. Use early return if it is truly needed.
```python
if len(node_set) < 2:
  return []
```

physlr/physlr.py
```diff
@@ -1021,107 +1020,118 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
-        return u, {v: i for i, vs in enumerate(communities) if len(vs) > 1 for v in vs}
+        return components
 
     @staticmethod
-    def determine_molecules_louvain(g, u):
-        "Apply louvain community detection algorithm after extracting bi-connected components."
-        import community as louvain
+    def detect_communities_k_clique(g, node_set, k=3):
+        """Apply k-clique community detection. Return communities."""
+        from networkx.algorithms import community as nxcommunity
```
```suggestion
```
Use `nx.community` instead.

physlr/physlr.py
afshinfard: If we remove these lines @sjackman we'll be calling the functions for single nodes... I added this line to avoid running the algorithms for single nodes. As we have partition might have single nodes.

sjackman: Since there are no k3 cliques if you have a single node or edge, the results should be identical without this line. If the results are identical, remove this line.

afshinfard: The results will be identical but I am worrying about calling k-cliques many times for nothing... and we can solve this issue just by checking a condition. So should we stick into summarizing the code or having less function calls @sjackman ?

physlr/physlr.py
afshinfard: I wanted to share this issue with you @sjackman .
If I use my own code, the result will be in this format:
`[{8, 9, 10, 11, 13}, {4, 5, 6, 7, 13}, {0, 1, 2, 3}]`
and if I use yours, and I used to do the thing the way you do, the result will be in this format:
```[frozenset({8, 9, 10, 11, 13}),
 frozenset({4, 5, 6, 7, 13}),
 frozenset({0, 1, 2, 3})]```
and we will have some issue calling some other functions. Not sure where, but I faced this issue and I decided to change it this way. Any other ideas on how to solve the issue?

File: physlr/physlr.py (hunk 8 above)
Comment (sjackman) [reply physlr/physlr.py]: Calling k-cliques with a single vertex or edge will exit very quickly. There's no need to optimize for this case. You can time it and test it empirically if you're concerned.

Comment (afshinfard) [reply physlr/physlr.py]: Just tried it to learn how to work with `timeit` :D
So here are the results:
```
>>>setup = '''
from networkx.algorithms import community as nxcommunity
import networkx as nx
node_set={1,2}
g=nx.Graph()
g.add_node(1)
g.add_node(2)
'''
>>>print(min(timeit.Timer('if len(node_set)>1: pass', setup=setup).repeat(7, 1000000)))
0.002583329100161791
>>>print(min(timeit.Timer('nxcommunity.k_clique_communities(g,3)', setup=setup).repeat(7, 1000000)))
0.04064908204600215
```
So the `if` condition is >15x faster but yep considering the calls for normal subgraphs, it's does not seem considerable.

File: physlr/physlr.py (hunk 14 above)
Comment (afshinfard) [reply physlr/physlr.py]: Right. changed.

File: physlr/physlr.py (hunk 9 above)
Comment (afshinfard) [reply physlr/physlr.py]: So should I let it remains as it is now @sjackman?

Comment (sjackman) [reply physlr/physlr.py]: I'd prefer to use `frozenset` wherever possible. Let's use the code I suggested unless you can point me to the particular error. When I know what the error is, we can figure out the best way to address it.

Comment (afshinfard) [reply physlr/physlr.py]: sure

Comment (afshinfard) [reply physlr/physlr.py]: It's working properly with your change :D

**Comment by afshinfard** [comment]
I made the changes and appreciate if you check.

**Comment by sjackman** [comment]
Use `nx.algorithms.community.kclique.k_clique_communities`. See for example PR https://github.com/bcgsc/physlr/pull/57

**Comment by afshinfard** [comment]
Ready for another round of review, @sjackman :D

File: physlr/physlr.py [hunk 17]
```diff
@@ -1189,106 +1189,109 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
...
     @staticmethod
-    def determine_molecules_biconnected_components(g, u):
-        "Separate bi-connected components."
-        cut_vertices = set(nx.articulation_points(g.subgraph(g.neighbors(u))))
-        components = list(nx.connected_components(g.subgraph(set(g.neighbors(u)) - cut_vertices)))
+    def detect_communities_biconnected_components(g, node_set):
+        """Separate bi-connected components. Return components."""
+        cut_vertices = set(nx.articulation_points(g.subgraph(node_set)))
+        components = list(nx.connected_components(g.subgraph(node_set - cut_vertices)))
         components.sort(key=len, reverse=True)
```
sjackman: Each `detect_communities_*` method needs to sort the communities by size.

sjackman: I suggest instead sorting just once in `determine_molecules` and removing all the other calls to `components.sort(key=len, reverse=True)`.

physlr/physlr.py
```diff
@@ -1300,6 +1303,8 @@ def determine_molecules_process(u):
 
     def physlr_molecules(self):
         "Separate barcodes into molecules."
+        if self.args.strategy not in [1, 2, 3, 4]:
```
sjackman: The list `[1,2,3,4]` is duplicated. Use `self.args.strategy not in strategy_switcher`

physlr/physlr.py
afshinfard: I am wondering about the purpose of using `components.sort(key=len, reverse=True)`. If it is not needed before splitting, then I can simply move it to the line just before the return line of `determine_molecules` function. Do you confirm?

sjackman: Yes, I agree. The purpose of sorting is so that the scaffold backbones output by `physlr backbone` are sorted by size.

physlr/physlr.py
```diff
@@ -1234,7 +1233,7 @@ def detect_communities_cosine_of_squared(g, node_set):
             sub_graph_copy = nx.Graph(g.subgraph(node_set))
             sub_graph_copy.remove_edges_from(edges_to_remove_barcode)
             cos_components = list(nx.connected_components(sub_graph_copy))
-            cos_components.sort(key=len, reverse=True)
+            # cos_components.sort(key=len, reverse=True)
```
sjackman: Please delete this line if it's not needed. Don't commit commented-out code.

afshinfard:
Results of using `k_clique` and `bi-connected` are identical. Pushing.

sjackman:
Thank you, Amirhossein!

afshinfard:
@sjackman you're welcome

PR #62, opened 2019-05-14
determine_molecules: partition+cluster+merge

afshinfard:


afshinfard:
@sjackman I was thinking about the issue of inconsistent results because of our random partitioning. What if we fix a seed for the random generator to a constant?

sjackman:
That would work if it were a single process, but molecule splitting is uses multiple processes. The random number seed needs to be reset for each individual barcode. I think that would work!

afshinfard:
I am thinking of merging this PR and then resolving the inconsistency issue later on. If you think it's fine.

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
                 for community in
                 Physlr.detect_communities_cosine_of_squared(g, bi_connected_component)]
 
+    @staticmethod
+    def partition_subgraph_into_bins_randomly(node_set, max_size=40):
+        """Prtition the subgraph into bins randomly for faster processing. Return bins."""
```
```suggestion
        """Partition the subgraph into bins randomly for faster processing. Return bins."""
```

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
                 for community in
                 Physlr.detect_communities_cosine_of_squared(g, bi_connected_component)]
 
+    @staticmethod
+    def partition_subgraph_into_bins_randomly(node_set, max_size=40):
```
```suggestion
    def partition_subgraph_into_bins_randomly(node_set, max_size=50):
```
Has this value been optimized, and found 40 to be best? If it's arbitrary and unoptimized, I prefer values along the 1-2-5, scale.

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
...
+        """Prtition the subgraph into bins randomly for faster processing. Return bins."""
+        bins_count = 1
+        if len(node_set) > max_size:
+            bins_count = 1 + int(len(node_set) / max_size)
```
sjackman: Replace 1338 to 1340 with
```suggestion
        bins_count = 1 + len(node_set) // max_size
```
Avoid conditionals where they're not necessary.

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
...
+        node_list = list(node_set)
+        random.shuffle(node_list)
+        size, leftover = divmod(len(node_set), bins_count)
+        bins = [node_list[0 + size * i: size * (i + 1)] for i in list(range(bins_count))]
```
```suggestion
        bins = [node_list[0 + size * i: size * (i + 1)] for i in range(bins_count)]
```
Do not convert to `list` where it's not necessary.

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
...
+        edge = size * bins_count
+        for i in list(range(leftover)):
```
```suggestion
        for i in range(leftover):
```
Do not convert to `list` where it's not necessary.

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
...
+            bins[i % bins_count].append(node_list[edge + i])
+        bin_sets = [set() for _ in range(len(bins))]
```
```suggestion
        bin_sets = [set() for _ in bins]
```

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
...
+        for i, c in zip(range(len(bins)), bins):
```
```suggestion
        for i, c in enumerate(bins):
```

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
...
+            bin_sets[i].update(set(c))
```
```suggestion
            bin_sets[i].update(c)
```
Is `set` needed here? Avoid unnecessary type conversions.

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
...
+            return communities
+        if strategy == 1:  # Merge by Initializing Louvain with the communities
+            return Physlr.detect_communities_louvain(g, node_set, communities)
+        # Ad-hoc Merge (default - strategy = 0)
+        merge_network = nx.Graph()
+        for i in range(len(communities)):
+            merge_network.add_node(i)
+        for i, com1 in enumerate(communities):
+            for k, com2 in enumerate(communities):
+                if i < k:
```
```suggestion
                if i >= j:
                    continue
```
and undent the remaining code.

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
...
+        if len(communities) == 1 and (node_set == 0 or strategy != 1):
+            return communities
+        if strategy == 1:  # Merge by Initializing Louvain with the communities
+            return Physlr.detect_communities_louvain(g, node_set, communities)
+        # Ad-hoc Merge (default - strategy = 0)
+        merge_network = nx.Graph()
+        for i in range(len(communities)):
+            merge_network.add_node(i)
+        for i, com1 in enumerate(communities):
+            for k, com2 in enumerate(communities):
```
```suggestion
            for j, com2 in enumerate(communities):
```
Use `j` if possible.

physlr/physlr.py
```diff
@@ -1332,6 +1332,79 @@ def determine_molecules_bc_cosine_of_squared(g, u):
...
+
+    @staticmethod
+    def determine_molecules_partition_split_merge(g, u):
+        """
+        Assign the neighbours of this vertex to molecules using fast heuristic community detection.
+        Pipeline: bi-connected (bc) + partition + bc + k-cliques communities + merge.
+        """
+        return [merged
+                for bi_connected_component in
+                Physlr.detect_communities_biconnected_components(g, set(g.neighbors(u)))
```
```suggestion
                Physlr.detect_communities_biconnected_components(g, g.neighbors(u))
```
Is this conversion necessary?

physlr/physlr.py
```diff
@@ -1345,6 +1418,8 @@ def determine_molecules(g, u, strategy):
             communities = Physlr.determine_molecules_bc_louvain(g, u)
         elif strategy == 4:  # bi-connected + sqCos
             communities = Physlr.determine_molecules_bc_cosine_of_squared(g, u)
+        elif strategy == 10:  # bi-connected + partition + bi-connected + k-cliques + merge
```
```suggestion
        elif strategy == 5:  # bi-connected + partition + bi-connected + k-cliques + merge
```
5?

physlr/physlr.py
```diff
@@ -1370,7 +1445,10 @@ def physlr_molecules(self):
                "Louvain community detection (after separating bi-connected components)",
             4: "\n\tStrategy: "
                "Community detection by cosine of squared adjacency matrix"
-               "(after separating bi-connected components)"
+               "(after separating bi-connected components)",
+            10: "\n\tStrategy: "
```
```suggestion
            5: "\n\tStrategy: "
```

physlr/physlr.py
sjackman:
```suggestion
        """
        Partition the subgraph into bins randomly for faster processing. Return bins.
        Warning: This function is not deterministic.
        """
```

physlr/physlr.py
afshinfard: umm I tested different values for a few examples and something between 30-40 seems to perform better. The lower the faster, and it is more likely to overcome some misassemblies by over-splitting. But if we set it to really low values then community detection sucks.

physlr/physlr.py
afshinfard: At first I set to `j` and `pylint` ouput error as I remember :D I'll check it again.

physlr/physlr.py
sjackman: Can you compare 40 to 50? If 40 performs better, I'm fine with that value.

physlr/physlr.py
afshinfard: um, I thought we might add more single-approach strategies later. So I reserved 5-9 for them. At the end of the day (the end of the year better applies here :)) ) we can change it to the a better number.

physlr/physlr.py
afshinfard: I have done that for fly. It's faster and not worst in terms of results quality  (theoretically might be better, but have not seen that for fly). So now do you prefer 40 or 50? I can test for human as well.

physlr/physlr.py
afshinfard: :laughing:

physlr/physlr.py
sjackman: Perhaps in a separate PR, I'd prefer to name the strategies rather than number them. Like `bic` and `k3` and `louvain`.

afshinfard: hum good idea.

physlr/physlr.py
afshinfard: Yes. The standard for all the functions that I am writing for `determine_molecules` is list of sets. Rather than consistency, It makes it so easy to build new pipelines for molecule separation based on these functions.

sjackman: `update` takes an enumeration as input, so the type that contains the enumeration doesn't affect the result. Both `bin_sets[i].update(set(c))` and `bin_sets[i].update(c)` yield the same result.

physlr/physlr.py
afshinfard: all `detect_communities_*` functions accept a node set. I am checking if I can change it to list of list... but on top of my head we needed it for mixing these functions. I might be wrong. I 'll check.

afshinfard: yes it's necessary.
we have `components = list(nx.connected_components(g.subgraph(node_set - cut_vertices)))` in `components = list(nx.connected_components(g.subgraph(node_set - cut_vertices)))` which requires `node_set` to be a set. We have kind of same thing in other com-det methods. Any alternatives for it?

physlr/physlr.py
afshinfard: `physlr/physlr.py:1370:16: C0103: Variable name "j" doesn't conform to snake_case naming style (invalid-name)` pylints error.

afshinfard: That's weird! a single-character word `j` does not conform to snake_case naming style while `i` and 'k' does!

physlr/physlr.py
afshinfard: Oh I see.. I did set `bin_sets = [set() for _ in bins]` so it is already a list of sets and... got it. Thanks

physlr/physlr.py
afshinfard: and there is another j used in another loop just after this loop, and pylint has not problem with that one :laughing:

physlr/physlr.py
sjackman: `g[u].keys()` may be more efficient than `set(g.neighbors(u))`. Want to time it?

physlr/physlr.py
sjackman: Just add `j` to `good-names` in `.pylintrc`. =)

physlr/physlr.py
sjackman: Do test zebrafish and human please.

afshinfard: I just decided to go with 50 for now and postpone this tuning to later. The difference is not that much and it's one less thing to worry about now :smile:

physlr/physlr.py
afshinfard: :+1:

afshinfard:
Any other changes required?

physlr/physlr.py
```diff
@@ -1357,6 +1357,81 @@ def determine_molecules_bc_cosine_of_squared(g, u):
...
+        random.shuffle(node_list)
+        size, leftover = divmod(len(node_set), bins_count)
+        bins = [node_list[0 + size * i: size * (i + 1)] for i in range(bins_count)]
+        edge = size * bins_count
+        for i in range(leftover):
+            bins[i % bins_count].append(node_list[edge + i])
+        bin_sets = [set() for _ in bins]
+        for i, c in enumerate(bins):
+            bin_sets[i].update(c)
+        return bin_sets
```
sjackman: Try…
```python
return [set(x) for x in bins]
```

physlr/physlr.py
afshinfard: woof really different:
```
print(min(timeit.Timer('a = g[u].keys()', setup=setup).repeat(7, 1000000)))
0.04027111700270325
print(min(timeit.Timer('a = set(g.neighbors(u))', setup=setup).repeat(7, 1000000)))
0.6511558880010853
```
tested it for a node with 12 neighbors.

physlr/physlr.py
afshinfard: changed and tested

physlr/physlr.py
afshinfard: changed and tested

sjackman: Nice! That's a win!

PR #63, opened 2019-05-23
Extensive separation

afshinfard:
using a combination of algorithms to perform the molecule separation

### DIFF
--- a/physlr/physlr.py
+++ b/physlr/physlr.py
@@ -1517,6 +1517,24 @@ class Physlr:
                     bi_connected_component, strategy=0)
                 ]
 
+    def determine_molecules_consensus(g, u):
+        """
+        Assign the neighbours of this vertex to molecules by
+        applying cosine of squared of the adjacency matrix for each bi-connected component.
+        """
+        return [community4
+                for bi_connected_component in
+                Physlr.detect_communities_biconnected_components(g, set(g.neighbors(u)))
+                for community in
+                Physlr.detect_communities_k_clique(g, bi_connected_component)
+                for community2 in
+                Physlr.detect_communities_cosine_of_squared(g, community, squaring=False)
+                for community3 in
+                Physlr.detect_communities_cosine_of_squared(g, community2)
+                for community4 in
+                Physlr.detect_communities_louvain(g, community3)
+                ]
+
     @staticmethod
     def determine_molecules(g, u, strategy):
         """Assign the neighbours of this vertex to molecules."""
@@ -1532,6 +1550,8 @@ class Physlr:
             communities = Physlr.determine_molecules_bc_cosine_of_squared(g, u)
         elif strategy == 5:  # bi-connected + partition + bi-connected + k-cliques + merge
             communities = Physlr.determine_molecules_partition_split_merge(g, u)
+        elif strategy == 20:  # Consensus Clustering
+            communities = Physlr.determine_molecules_consensus(g, u)
         else:
             exit("\033[93m Wrong input argument: --separation-strategy!\033[0m")
 
--- a/physlr/physlr.py
+++ b/physlr/physlr.py
@@ -1523,24 +1523,6 @@ class Physlr:
         Assign the neighbours of this vertex to molecules
         by Applying a queue of different algorithms on top of each other.
         """
-        # communities = []
-        # communities2 = []
-        # for bi_connected_component \
-        #         in Physlr.detect_communities_biconnected_components(g, set(g.neighbors(u))):
-        #     communities.extend(Physlr.detect_communities_k_clique(g, bi_connected_component))
-        # for community in communities:
-        #     communities2.extend(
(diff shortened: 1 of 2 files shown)


afshinfard:
I am waiting for the results to confirm that this latest version of the code produces the same results as what I've shown in the last meetings.

afshinfard:
I just changed `max-public-methods` from 100 to 150 but I have an idea to make the code related to molecule separation more compact with less methods, and I will reduce it in another PR when I reduce the code size.

physlr/physlr.py
```diff
@@ -1517,6 +1517,45 @@ def determine_molecules_partition_split_merge(g, u):
...
+        Assign the neighbours of this vertex to molecules
+        by Applying a queue of different algorithms on top of each other.
+        """
+        from collections import deque
+
+        communities = [g[u].keys()]
+        communities_final = []
+
+        alg_list = deque(["bc", "k3", "bc", "cos", "bc", "sqCos", "bc"])
+        while alg_list:
```
```suggestion
        for algorithm in ["bc", "k3", "bc", "cos", "bc", "sqCos", "bc"]:
```

physlr/physlr.py
```diff
@@ -1517,6 +1517,45 @@ def determine_molecules_partition_split_merge(g, u):
                     bi_connected_component, strategy=0)
                 ]
 
+    @staticmethod
+    def determine_molecules_extensive(g, u):
+        """
+        Assign the neighbours of this vertex to molecules
+        by Applying a queue of different algorithms on top of each other.
+        """
+        from collections import deque
```
```suggestion
```

physlr/physlr.py
```diff
@@ -1517,6 +1517,45 @@ def determine_molecules_partition_split_merge(g, u):
...
+
+        communities = [g[u].keys()]
+        communities_final = []
+
+        alg_list = deque(["bc", "k3", "bc", "cos", "bc", "sqCos", "bc"])
+        while alg_list:
+            communities_final.clear()
+            algoritm = alg_list.popleft()
```
```suggestion
```

physlr/physlr.py
```diff
@@ -1517,6 +1517,45 @@ def determine_molecules_partition_split_merge(g, u):
...
+                            g, component, squaring=False, threshold=0.4))
+            elif algoritm == "sqCos":
+                for component in communities:
+                    communities_final.extend(
+                        Physlr.detect_communities_cosine_of_squared(g, component))
+            elif algoritm == "louvain":
+                for component in communities:
+                    communities_final.extend(
+                        Physlr.detect_communities_louvain(g, component))
+            communities = communities_final.copy()
```
```suggestion
            communities = communities_final
```
What's the purpose of the `copy` here?

physlr/physlr.py
sjackman: Have you tested does having the `bc` between each algorithm have an effect? Does it give better separation than say `["bc", "k3", "cos", "sqCos"]`?

afshinfard: Nope :D I'll test but Inanc suggested that since bc is extremely faster than the others it wouldn't be a pain to do so.

physlr/physlr.py
afshinfard: when I set `communities = communities_final` it seems that they become tied to each other, and int the next iteration when I do `communities_final.clear()` also `communities` become cleared and the function do not work as expected.

physlr/physlr.py
afshinfard: I have restored the code to `["bc", "k3", "cos", "sqCos"]` for now, and I will test idea of multiple bcs later.

physlr/physlr.py
afshinfard: So if its fine, the PR is ready for another review.

sjackman: In that case use `communities_final = []` rather than `communities_final.clear()` and remove the `.copy()` here.

physlr/physlr.py
```diff
@@ -1517,6 +1517,41 @@ def determine_molecules_partition_split_merge(g, u):
...
+        return communities_final
```
```suggestion
        return communities
```

sjackman:
```
************* Module physlr.physlr
physlr/physlr.py:1531:15: E0602: Undefined variable 'algoritm' (undefined-variable)
physlr/physlr.py:1535:17: E0602: Undefined variable 'algoritm' (undefined-variable)
physlr/physlr.py:1539:17: E0602: Undefined variable 'algoritm' (undefined-variable)
physlr/physlr.py:1544:17: E0602: Undefined variable 'algoritm' (undefined-variable)
physlr/physlr.py:1548:17: E0602: Undefined variable 'algoritm' (undefined-variable)
physlr/physlr.py:1529:12: W0612: Unused variable 'algorithm' (unused-variable)
```
`algoritm` should be `algorithm`.

physlr/physlr.py
afshinfard: Thank Shaun

sjackman:
Thanks, Amirhossein!

PR #65, opened 2019-06-12
Fix an edge case bug

schutzekatze:


sjackman:
How does this bug affect existing runs?

schutzekatze:
It shouldn't affect them the output of any runs.  It was an edge case in which input worker, after arriving at EOF, still ends up writing to its buffer (an empty block) and messes up the read counter, causing an assertion to fail.

lcoombe:
This was the error message I was seeing, just to have it documented:
```physlr-indexlr: ./include/indexlr-workers.h:293: void MinimizeWorker::work(): Assertion `reads.dataCounter - 1 < sizeof(reads.data) / sizeof(reads.data[0])' failed.```

schutzekatze:
Tested with chr4 fly reads, output is the same as before the pull request.

lcoombe:
Thank you for the quick fix, @schutzekatze!

PR #70, opened 2019-08-13
determine_molecules: bug fix

afshinfard:
fixed a bug

### DIFF
--- a/data/Makefile
+++ b/data/Makefile
@@ -14,7 +14,7 @@ physlr_path=$(shell dirname $$(dirname $(realpath $(MAKEFILE_LIST))))
 bin=$(physlr_path)/bin
 
 # Python interpreter.
-python=env PYTHONPATH=$(physlr_path) pypy3
+python=env PYTHONPATH=$(physlr_path) /projects/btl/aafshinfard/virtuEnv/pypy3/bin/pypy3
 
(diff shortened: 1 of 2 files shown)


afshinfard:
The failed check seem to be for `physlr_paf_metrics()` (not in my edits for the bug fix) and caused by unused variables in that function. It's a bit tricky and actually those variables are kinda used. They seemingly need to bring AI into `pylint` to distinguish such cases :))

lcoombe:
@afshinfard - I just fixed the pylint error in #71 :)

When we have that merged into master, you could rebase your branch on master and then azure should be happy.

afshinfard:
Thanks @lcoombe, I'll rebase tomorrow.

data/Makefile
```diff
@@ -14,7 +14,7 @@ physlr_path=$(shell dirname $$(dirname $(realpath $(MAKEFILE_LIST))))
 bin=$(physlr_path)/bin
 
 # Python interpreter.
-python=env PYTHONPATH=$(physlr_path) pypy3
+python=env PYTHONPATH=$(physlr_path) /projects/btl/aafshinfard/virtuEnv/pypy3/bin/pypy3
```
sjackman: Please revert this line.

PR #71, opened 2019-08-13
Fix pylint errors in physlr_paf_metrics

lcoombe:
Fix 'unused variable' errors from pylint.


afshinfard:
Hmm, I did not know the trick of using an underscore for those variables. and yeah am new in Python I will definitely face more like this :D

PR #72, opened 2019-08-15
Physlr: compute intersection of neighbour minimizers as edge weight

jowong4:
Can i get some reviews on the neighbour min code i wrote

### DIFF
--- a/physlr/physlr.py
+++ b/physlr/physlr.py
@@ -1218,7 +1218,7 @@ class Physlr:
         Physlr.neighbour_min = neighbour_min
         with multiprocessing.Pool(self.args.threads) as pool:
             edges2 = dict(x for l in pool.map(self.physlr_overlap_minimizer_intersection, progress(edge_list),
-                                            chunksize=100) for x in l)
+                                              chunksize=100) for x in l)
         edges = edges2
         print(int(timeit.default_timer() - t0), "Recalculated", len(edges), "edges weights", file=sys.stderr)
 
--- a/physlr/physlr.py
+++ b/physlr/physlr.py
@@ -1184,6 +1184,7 @@ class Physlr:
 
     @staticmethod
     def physlr_overlap_minimizer_intersection(edge):
+        "Calculate the intersection of minimizers neighbours of each edge node"
         return [(edge, len(Physlr.neighbour_min[edge[0]] & Physlr.neighbour_min[edge[1]]))]
 
     def physlr_overlap(self):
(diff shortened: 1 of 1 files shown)


physlr/physlr.py
```diff
@@ -1201,6 +1207,27 @@ def physlr_overlap(self):
             (u, v) for bxs in progress(mxtobxs.values()) for u, v in itertools.combinations(bxs, 2))
         print(int(timeit.default_timer() - t0), "Loaded", len(edges), "edges", file=sys.stderr)
 
+        edge_list = list(edges)
```
lcoombe: Couldn't you just iterate over the `edges` counter object instead of doing this explicit conversion?

physlr/physlr.py
```diff
@@ -1201,6 +1207,27 @@ def physlr_overlap(self):
...
+        neighbour_min = defaultdict(list)
+
+        for edge in edge_list:
```
```suggestion
        for u, v in edges:
```

physlr/physlr.py
```diff
@@ -1201,6 +1207,27 @@ def physlr_overlap(self):
...
+            neighbour_min[edge[0]] += bxtomxs[edge[1]]
+            neighbour_min[edge[1]] += bxtomxs[edge[0]]
```
lcoombe: What about doing the recalculation of the edge weight after adding the edges to the graph (line 1231)? Then, you could just something like `bxtomxs[v] for v in g.neighbours(bx)` to find the minimizers of the neighbours of a given node

physlr/physlr.py
```diff
@@ -1201,6 +1207,27 @@ def physlr_overlap(self):
...
+        with multiprocessing.Pool(self.args.threads) as pool:
+            edges2 = dict(x for l in pool.map(self.physlr_overlap_minimizer_intersection,
+                                              progress(edge_list),
+                                              chunksize=100) for x in l)
+        edges = edges2
+        print(
+            int(timeit.default_timer() - t0),
+            "Recalculated", len(edges), "edges weights", file=sys.stderr)
+
         for (u, v), n in progress(edges.items()):
```
lcoombe: So currently we are not doing the update of the edge weights on the graph right?

physlr/physlr.py
jowong4: Yeah that is the local version. I wasn't sure all the intricacies would work out so I converted it to a list first. The local copy uses `edges`

physlr/physlr.py
jowong4: `(u, v)` in the local version

physlr/physlr.py
jowong4: It made more sense to me to do update the edge weight and add the edges to the graph

physlr/physlr.py
jowong4: well I can benchmark that later and see the difference in time

physlr/physlr.py
lcoombe: But you aren't updating `edges` variable until the very end, right? I don't quite understand what you're saying..

jowong4: By local, I meant the local copy of the code on my pc. I have implemented it like you suggested.

PR #75, opened 2019-09-11
physlr-filter-bxmx.cc: calculate high whisker correctly

jowong4:
fix high whisker calculation


sjackman:
The previous code was correct. The upper whisker is Q3 + 1.5 * IQR. See https://www.rdocumentation.org/packages/grDevices/versions/3.6.1/topics/boxplot.stats

PR #76, opened 2019-09-13
determine_molecules: bug fix

afshinfard:


afshinfard:
Fixed a bug (missing loop) for binning nodes for faster separation.

PR #78, opened 2019-09-19
Option to filter minimizer with bloomfilter

jowong4:
Deletion of `assert_good`  and the inclusion of `IOUtil.h` is due to compilation error since there is an assert_good function in the `btl_bloomfilter`.

Currently, the bloomfilter is a exclusion filter where if kmer `ATCG` is in the bloomfilter, it will not be a minimizer. I plan to add another flag `-mode` where we can exclude or include kmer in the bloomfilter


src/include/indexlr-minimize.h
```diff
@@ -33,6 +35,22 @@ struct HashData
 	char strand;
 };
 
+struct find_HashData
```
schutzekatze: Seems redundant, should be merged with HashData.

src/include/indexlr-minimize.h
```diff
@@ -112,14 +130,51 @@ getMinimizers(const HashValues& hashes, const unsigned w)
 	return minimizers;
 }
 
-// Test the condition of a I/O stream.
-static inline void
-assert_good(const std::ios& stream, const std::string& path)
+static inline HashValues
+getMinimizers(const HashValues& hashes, const unsigned w, const BloomFilter& bloomFilter)
```
schutzekatze: There seems to be a lot of code in common between this and the other `getMinimizers`. Can you separate that functionality in another function?

src/physlr-indexlr.cc
```diff
@@ -26,9 +27,11 @@ minimizeReads(
     const size_t k,
     const size_t w,
     const size_t t,
+    const bool withBloomFilter,
     const bool withPositions,
     const bool withStrands,
-    const bool verbose)
+    const bool verbose,
+    BloomFilter& bloomFilter)
```
schutzekatze:
```suggestion
    const BloomFilter& bloomFilter)
```

src/physlr-indexlr.cc
```diff
@@ -119,6 +135,15 @@ main(int argc, char* argv[])
 				          << ": Using more than 5 threads does not scale, reverting to 5.\n";
 			}
 			break;
+		case 'b':
+			withBloomFilter = true;
+			std::cerr << "loading bloomfilter from " << optarg << std::endl;
```
schutzekatze:
```suggestion
			std::cerr << "Loading bloomfilter from " << optarg << std::endl;
```

src/include/indexlr-minimize.h
jowong4: Removed the code as new implementation no longer requires it

src/include/indexlr-minimize.h
jowong4: Removed with new implementation

src/physlr-indexlr.cc
jowong4: Fixed

src/physlr-indexlr.cc
jowong4: fixed

jowong4:
Any suggestions for alternative names of rBloomFilter and sBloomFilter. They stand for repetitive/ repeats Bloom filter and solid Bloom filter

data/Makefile
```diff
@@ -642,11 +642,35 @@ minsize=2000
 
 # Index a FASTA file.
 %.k$k-w$w.physlr.tsv: %.fa
+ifeq ($(r),)
+ifeq ($(s),)
```
lcoombe: Is there a default value for `r` and `s` in the Makefile? I'm thinking if at least there is a default setting we can add a comment to know what that variable means

src/physlr-indexlr.cc
```diff
@@ -59,16 +77,18 @@ static void
...
-	             "  --strand   include minimizer strand in the output\n"
-	             "  -v         enable verbose output\n"
-	             "  -o FILE    write output to FILE, default is stdout\n"
-	             "  -t N       use N number of threads (default 1, max 5)\n"
-	             "  --help     display this help and exit\n"
-	             "  FILE       space separated list of FASTQ files\n";
+	          << "  -k K -w W [-r repeat_bf_path] [-s solid_bf_path] [-v] [-o FILE] FILE...\n\n"
+	             "  -k K        use K as k-mer size\n"
+	             "  -w W        use W as sliding-window size\n"
+	             "  -r repeat_bf_path  use a bloomfilter to filter out repetitive minimizers\n"
```
lcoombe:
```suggestion
	             "  -r repeat_bf_path  use a Bloom filter to filter out repetitive minimizers\n"
```

src/physlr-indexlr.cc
```diff
@@ -59,16 +77,18 @@ static void
...
+	             "  -s solid_bf_path  use a bloomfilter to filter in solid minimizers\n"
```
lcoombe:
```suggestion
	             "  -s solid_bf_path  use a Bloom filter only select solid minimizers\n"
```
Just thinking 'filter in' doesn't sound right

src/vendor/btl_bloomfilter/pythonInterface/python_wrapper.readme
```diff
@@ -0,0 +1,21 @@
+Python wrapper for BloomFilter.hpp
```
lcoombe: Do we need to be including all of these interface files, etc. in this commit?

src/physlr-indexlr.cc
```diff
@@ -4,9 +4,11 @@
 // Originally written for Physlr: (https://github.com/bcgsc/physlri
```
lcoombe:
```suggestion
// Originally written for Physlr: (https://github.com/bcgsc/physlr)
```

src/physlr-indexlr.cc
```diff
@@ -146,6 +192,9 @@ main(int argc, char* argv[])
 	} else if (infiles.empty()) {
 		printErrorMsg(progname, "missing file operand");
 		failed = true;
+	} else if (withRepeat && withSolid) {
```
lcoombe: Can you clarify what type of error this statement is checking for?

jowong4: That was something I left in when I first wrote the code. Wasn't sure how to approach using both Bloom filter. I will remove this line

src/vendor/btl_bloomfilter/pythonInterface/python_wrapper.readme
jowong4: The intention was to just unzip the vendored repo and forget about it. Currently, abyss is like that too

data/Makefile
jowong4: I've added a comment and made the variables more verbose so people should understand the intention

src/vendor/btl_bloomfilter/pythonInterface/python_wrapper.readme
lcoombe: Ok makes sense!

PR #79, opened 2019-09-27
Pylint errors

jowong4:
Pylint got updated recently. New checks came with the new updates and all our future CIs will break as a result.

Added exception to non top level import.
Changed all `exit` to `sys.exit`

### DIFF
--- a/physlr/physlr.py
+++ b/physlr/physlr.py
@@ -1348,7 +1348,8 @@ class Physlr:
     def physlr_split_minimizers(self):
         "Given the molecule overlap graph, split the minimizers into molecules"
         if len(self.args.FILES) < 2:
-            sys.exit("physlr split-minimizers: error: graph file and bx to minimizer inputs required")
+            exit_message = "physlr split-minimizers: error: graph file and bx to minimizer inputs required"
+            sys.exit(exit_message)
         g = self.read_graph([self.args.FILES[0]])
         bxtomxs = self.read_minimizers([self.args.FILES[1]])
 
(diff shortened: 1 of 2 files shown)


physlr/physlr.py
```diff
@@ -1500,9 +1506,6 @@ def detect_communities_cosine_of_squared(g, node_set, squaring=True, threshold=0
         Square the adjacency matrix and then use cosine similarity to detect communities.
         Return communities.
         """
-        import scipy as sp
```
lcoombe: I guess the good thing about having the import statements like this was that the user wouldn't require those packages unless they were using this specific molecule separation mode?

What was the pylint error here? What do we think about following the error vs. editing the config file to ignore such issues?

jowong4:
```
physlr/physlr.py:1492:8: C0415: Import outside toplevel (community) (import-outside-toplevel)
physlr/physlr.py:1503:8: C0415: Import outside toplevel (scipy) (import-outside-toplevel)
physlr/physlr.py:1504:8: C0415: Import outside toplevel (numpy) (import-outside-toplevel)
physlr/physlr.py:1811:8: C0415: Import outside toplevel (physlr.mkt) (import-outside-toplevel)
physlr/physlr.py:1812:8: C0415: Import outside toplevel (numpy) (import-outside-toplevel)
```

jowong4: might able to disable that warning specifically. I dont know

lcoombe: If we want to, you can add an exception in the `.pylintrc`

jowong4: Done

PR #81, opened 2019-10-03
Fix gap.bed error in data/Makefile

janetxinli:


jowong4:
Remind me, @lcoombe , are we trying to trigger this rule if 

```
%.fa.gap.bed: %.fa
	$(python) $(bin)/physlr fasta-gaps $< >$@
```
if this is missing `$(name)/$(ref).fa.gap.be`

lcoombe:
> Remind me, @lcoombe , are we trying to trigger this rule if
> 
> ```
> %.fa.gap.bed: %.fa
> 	$(python) $(bin)/physlr fasta-gaps $< >$@
> ```
> 
> if this is missing `$(name)/$(ref).fa.gap.bed`?

Yup that's right -- We want to make sure the `$(name)/$(ref).fa.gap.bed` file is made before the `plotpaf.rmd` script is executed.

lcoombe:
Thanks Janet! :+1:

PR #82, opened 2019-10-04
Fix a rare deadlock :D

schutzekatze:


lcoombe:
Thanks Vlad!

src/include/indexlr-buffer.h
```diff
@@ -29,7 +29,7 @@ struct Slot
 	std::mutex busy;
 	bool occupied = false;
 	std::condition_variable occupancyChanged;
-	size_t lastTenant = 0; // Required to ensure read order
+	int64_t lastTenant = -1; // Required to ensure read order
```
sjackman: @schutzekatze There's also `ssize_t` if you were looking for a signed size type.

schutzekatze: Ah, the reason I picked int64_t is because, initially, I put long there, but clang-tidy issued a warning and suggest int64 instead.

PR #84, opened 2019-10-09
separate --prune and --prune-bridges

afshinfard:
--prune has two functionalities:
- prune branches smaller than this size.
- prune bridges smaller than this.

While the first one is used for simplifying the graph, the second is a heuristic for miss-assembly correction; so I though we need to separate them (although they still have the same value!) so I added `prune_bridges`
I am currently troubleshooting this heuristic way of correction which is causing unwanted cuts and joins; we can then modify the value in the makefile for each one or, we may end up changing the way we do the correction.
[update]: added a 3rd --prune being --prune-junctions (was --min-branch before, for splitting junctions in an MST)

### DIFF
--- a/data/Makefile
+++ b/data/Makefile
@@ -690,7 +690,7 @@ minsize=2000
 
 # Determine the maximum spanning tree.
 %.mst.tsv: %.tsv
-	$(python) $(bin)/physlr mst -V$V --prune=10 --prune-bridges=10 $< >$@
+	$(python) $(bin)/physlr mst -V$V --prune-branches=10 --prune-bridges=10 $< >$@
 
 # Separate a graph into its biconnected components by removing its cut vertices.
 %.bic.tsv: %.tsv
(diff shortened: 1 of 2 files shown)


data/Makefile
```diff
@@ -690,20 +690,20 @@ minsize=2000
 
 # Determine the maximum spanning tree.
 %.mst.tsv: %.tsv
-	$(python) $(bin)/physlr mst -V$V --prune=10 $< >$@
+	$(python) $(bin)/physlr mst -V$V --prune=10 --prune-bridges=10 $< >$@
```
lcoombe: If `--prune` is for branches, and `--prune-bridges` is for bridges, should we also rename `--prune` to `--prune-branches` to make its purpose clear?

lcoombe: Also, should we we add these values as variables in the Makefile so they are easier to change?

afshinfard: makes sense! I'll just do it.

afshinfard: @lcoombe I was checking the Makefile and I found that most of the params are not defined as variables. Also, we use two different values of `prune-branches` one for `backbone` (0 now) and the other for `backbone-graph` (10 now). Should I define 2 different Makefile variables for these two stages? alternatively I can embed a fixed value (0) for `backbone`.
Or I can forget about adding Makefile variables for them. Adding them makes my troubleshooting easier but ... yeah

lcoombe: I'd say if values are hard-coded right now (ie we won't change the values usually), it would be fine to keep as is, but if you will be changing the values I think it's worth adding to the Makefile vs. having to go in and edit the value each time.
So yeah, we could add variables as we do PRs like this -- so only make the values variables when we realize we want to change them?

afshinfard:
[update]: added a 3rd --prune being --prune-junctions (was --min-branch before, for splitting junctions in an MST)

data/Makefile
afshinfard: So I am done with `--prune-branches`, and we only need to decide about the Makefile

lcoombe: I'm fine with it either way -- Just thought it would be easier for you to troubleshoot if you're going to be changing that parameter, but it's totally up to you!

afshinfard: sorry it did not show me your recent comment. Yeah I think it's good to stay with harcoded version for now; I don't think we will need to change it for different runs. So no more changes and I just added a 3rd one: `--prune-junctions`

lcoombe: Sounds good!

PR #86, opened 2019-10-11
physlr_remove_bridges_graph

afshinfard:
takes the overlap graph, removes the bridges in its MST(s) iteratively and output the bridge-removed overlap graph (not the backbone graph).
Purpose: troubleshooting the current missassembly detection method and developing a new method of correction.

### DIFF
--- a/data/Makefile
+++ b/data/Makefile
@@ -698,7 +698,7 @@ minsize=2000
 
 # Determine the bridge-removed graph from the overlap TSV.
 %.bridge-removed.tsv: %.tsv
-	$(time) $(python) $(bin)/physlr bridge-removed --prune-branches=10 --prune-bridges=10 -s0 $< >$@
+	$(time) $(python) $(bin)/physlr remove-bridges-graph --prune-branches=10 --prune-bridges=10 -s0 $< >$@
 
 # Determine the backbone graph from the overlap TSV.
 %.backbone.tsv: %.tsv
--- a/data/Makefile
+++ b/data/Makefile
@@ -696,6 +696,10 @@ minsize=2000
 %.bic.tsv: %.tsv
 	$(python) $(bin)/physlr biconnected-components $< >$@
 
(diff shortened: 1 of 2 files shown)


afshinfard:
@lcoombe I am still not sure if I should PR to master or not, the main purpose is troubleshooting (that may take some time) but like some other steps it take a long time to do this single step of removing bridges and I see it as a good checkpoint.

data/Makefile
```diff
@@ -696,6 +696,10 @@ minsize=2000
 %.bic.tsv: %.tsv
 	$(python) $(bin)/physlr biconnected-components $< >$@
 
+# Determine the bridge-removed graph from the overlap TSV.
+%.bridge-removed.tsv: %.tsv
+	$(time) $(python) $(bin)/physlr bridge-removed --prune-branches=10 --prune-bridges=10 -s0 $< >$@
```
```suggestion
	$(time) $(python) $(bin)/physlr bridge-removal --prune-branches=10 --prune-bridges=10 -s0 $< >$@
```
I know this is a small suggestion -- I just wonder if that is a slightly more clear name

afshinfard: yep I agree, especially as the `-` that grammatically (if am right) is needed in between is now hidden behind the common dash we use.

afshinfard: am, just a small concern: 
Other functions are named with the type of data they output. For example: `physlr backbone`, `physlr backbone-graph`, `physlr mst`. and in this example we have a bridge-removed graph and that is why I used this naming... while bridge-removal is the action (?). I myself am still feel bad with bridge-removed but do you have alternatives that match name of other functions?
update: we do have functions like `physlr count-molecules`. so `physlr bridge-removal` or `physlr remove-bridges`?

lcoombe: Yeah I like `remove-bridges`

afshinfard:
and I had a test run to see if the function works. I simply ran the backbone stage on the output of this new checkpoint, and the backbone stage should remove no more bridges. and it did not.

PR #88, opened 2019-10-17
data: Download the fly genome from Ensembl

sjackman:
The Ensembl genome has more intuitive chromosome names.
2L 2R 3L 3R 4 X Y MT


sjackman:
This PR doesn't change the default fly reference genome from NCBI to Ensembl, but that should be done once the on-disk analyses are updated.

data/Makefile
```diff
@@ -14,7 +14,8 @@ physlr_path=$(shell dirname $$(dirname $(realpath $(MAKEFILE_LIST))))
 bin=$(physlr_path)/bin
 
 # Python interpreter.
-python=env PYTHONPATH=$(physlr_path) pypy3
+#xxx python=env PYTHONPATH=$(physlr_path) pypy3
+python=env PYTHONPATH=$(physlr_path) python3
```
afshinfard: Is that intentional?

sjackman: Drat. No. Thanks for catching that.

PR #90, opened 2019-10-17
read_tsv: accept graphs with no edge

afshinfard:
I was using graph format to input a list of barcodes, and I faced a "bug" in `read_tsv()`. If your graph has no edges the `read_tsv` method outputs an error message and exit. I cannot call it bug as we usually do not like graphs with no edge, but If we use this for specific usages in which graphs with no edge makes sense (like what I am doing now - to input a list of barcodes) then we need to fix this.
What do you think?

### DIFF
--- a/physlr/physlr.py
+++ b/physlr/physlr.py
@@ -202,6 +202,8 @@ class Physlr:
                         print("Unexpected header:", line, file=sys.stderr)
                         sys.exit(1)
                     line = fin.readline()
+                    if not line: # a graph with no edges
+                        break
                     if Physlr.args.verbose >= 2:
                         progressbar.update(len(line))
                 xs = line.split()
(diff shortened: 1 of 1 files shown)


jowong4:
or something like this?

`def read_tsv(g, filename, type="graph"):` so we tell the function what type of tsv it should expect?

afshinfard:
> or something like this?
> 
> `def read_tsv(g, filename, type="graph"):` so we tell the function what type of tsv it should expect?

even if it expects a graph, the graph might have no edges in it; and the current code stop the job; which makes sense for current graphs we have. but if you consider any graph then it's a bug.
But that's a good suggestion. I can do that as well, and leave the graph reading part as is.

jowong4:
ah I see your point. A graph with no edges is still a graph and it breaks if the graph is like that. Not sure, if this is necessary but I would also add a logging message. Something like "Warning graph has no edges".

PR #93, opened 2019-10-22
Parse stLFR reads

jowong4:


sjackman:
> No description provided.

Please fill out the description. For example:

> An example of a STLFR read ID is…
> ```
> CL100050407L1C001R001_1#224_1078_917/1
> ```
> The barcode is composed of the characters between the `#` and `/` characters. It is composed of digits and underscores (three numbers separated by two underscores).

Here's a Makefile snippet that I used to convert the STLFR read IDs to a `BX:Z:` tagged FASTQ file.

```Makefile
# Interleave the reads, and add the BX tag.                                     
stlfr1.fq.gz: split_read.1.fq.gz split_read.2.fq.gz                             
    seqtk mergepe $^ | gsed '1~4{s~/.*~~;s/#/ BX:Z:/;}' | $(gzip) >$@           
```
`/projects/btl/datasets/hsapiens/NA12878/stLFR/Makefile`

Here's the result FASTQ file:
`/projects/btl/datasets/hsapiens/NA12878/stLFR/stlfr1.fq.gz`
For example:
```console
$ gunzip -c split_read.1.fq.gz | head -n1
@CL100050407L1C001R001_1#224_1078_917/1	1	1
$ gunzip -c stlfr1.fq.gz | head -n1
@CL100050407L1C001R001_1 BX:Z:224_1078_917
```

`string::find` and `string::rfind` can be used to find the `#` and `/` characters. A regex is overkill for that purpose.
`string::substr` can be used to extract the barcode found between the `#` and `/` characters, and will be more efficient than `std::regex_replace`.

It's usually best to avoid new command line options like `--stlfr` that specify the input data type and simply autodetect the input data type, when it's possible. For example, if the FASTQ header does not contain `BX:Z` and the read ID contains `#`, then you can assume that it's an STLFR read.

sjackman:
As a rule of thumb, regexes are usually not the optimal solution in C++.

PR #94, opened 2019-10-23
Auto detect stLFR

jowong4:
As per @sjackman comments in #93 , I have added code to auto detect stLFR format instead of having the user specify the data type with `--stlfr`. The code will also now work with a mixed data set should a user want to combine stLFR and 10X data

Reminder:
10X format:
`@HISEQX1:42:H352FALXX:5:1119:3397:59446_AAAGTAGCATCTTAGG/1	BX:Z:AAAGTAGCATCTTAGG-1
` where `AAAGTAGCATCTTAGG-1` is the barcode

stLFR format:
`@V10002828L1C001R013000000#543_288_92/1 1       1` where `543_288_92` is the barcode


src/include/indexlr-workers.h
```diff
@@ -338,16 +332,22 @@ MinimizeWorker::work()
 
 			auto minimizers = getMinimizers(hashes, w);
 
-			if (stlfr) {
-				std::regex right("^.*#+");
-				std::regex left("/.*");
-
-				read.barcode = std::regex_replace(read.id, right, "");
-				read.barcode = std::regex_replace(read.barcode, left, "");
+			if (read.id.find("BX:Z:") == std::string::npos) {
```
sjackman:
```suggestion
			if (read.barcode.empty()) {
```

src/include/indexlr-workers.h
```diff
@@ -338,16 +332,22 @@ MinimizeWorker::work()
...
+				size_t rightPos = read.id.find('#');
+				if (rightPos == std::string::npos) {
+					ss << read.barcode;
```
sjackman: `read.barcode` should be empty here. What's the purpose of this line?

src/include/indexlr-workers.h
```diff
@@ -338,16 +332,22 @@ MinimizeWorker::work()
...
+				} else {
+					size_t leftPos = read.id.find('/', rightPos + 1);
```
sjackman:
```suggestion
					size_t leftPos = read.id.rfind('/');
```

src/include/indexlr-workers.h
```diff
@@ -338,16 +332,22 @@ MinimizeWorker::work()
...
+					if (leftPos == std::string::npos) {
+						ss << read.barcode;
```
sjackman: `read.barcode` should be empty here. What's the purpose of this line?

src/include/indexlr-workers.h
```diff
@@ -338,16 +332,22 @@ MinimizeWorker::work()
...
 			auto minimizers = getMinimizers(hashes, w);
 
-			if (stlfr) {
-				std::regex right("^.*#+");
-				std::regex left("/.*");
-
-				read.barcode = std::regex_replace(read.id, right, "");
-				read.barcode = std::regex_replace(read.barcode, left, "");
+			if (read.id.find("BX:Z:") == std::string::npos) {
+				size_t rightPos = read.id.find('#');
```
sjackman: Should `rightPos` be `leftPos`? Better to call it `sharpPos`.

src/include/indexlr-workers.h
sjackman: Shouldn't `leftPos` be `rightPos`? Better to call it `slashPos`.

src/include/indexlr-workers.h
```diff
@@ -338,16 +332,22 @@ MinimizeWorker::work()
...
+				if (rightPos == std::string::npos) {
+					ss << read.barcode;
+				} else {
+					size_t leftPos = read.id.find('/', rightPos + 1);
+					if (leftPos == std::string::npos) {
+						ss << read.barcode;
+					} else {
+						ss << read.id.substr(rightPos + 1, leftPos - 1 - rightPos);
```
sjackman: Before this line assert
```c++
assert(leftPos < rightPos);
```

src/include/indexlr-workers.h
jowong4: Yeah, mixed up left and right.

src/include/indexlr-workers.h
jowong4: Before I implemented this, `ss << read.barcode` will execute regardless of it the barcode is empty so I thought I should keep that

src/include/indexlr-workers.h
jowong4: Yeah, mixed up left and right. I was wondering what is better a name to call that variable. Thanks for the suggestion

src/include/indexlr-workers.h
jowong4: Before I implemented this, `ss << read.barcode` will execute regardless of if the barcode is empty so I thought I should keep that

src/include/indexlr-workers.h
jowong4: And, I didn't want to use `rfind` because that gives me the last occurance, and if there are optional comments with forward slash, that would not work

src/include/indexlr-workers.h
jowong4: `read.barcode` is not empty. It will contain `NA`.

jowong4: so I could do `read.id.find("BX:Z:") == std::string::npos` or `read.barcode=="NA"`

src/include/indexlr-workers.h
jowong4: The line of code is necessary in order to output NA for reads that have no barcode in the read

src/include/indexlr-workers.h
jowong4: The line of code is necessary in order to output NA for reads that have no barcode in the read

src/include/indexlr-workers.h
sjackman:
```suggestion
			if (read.barcode == "NA") {
```

src/include/indexlr-workers.h
sjackman: This logic of detecting the stLFR barcode should be moved up to where the `barcode` member is initialized at
https://github.com/bcgsc/physlr/blob/f9665f5392fe70068c95318cb06efe32f5f240ed/src/include/indexlr-workers.h#L286

src/include/indexlr-workers.h
sjackman: I believe (but you'll want to confirm) that `read.id` doesn't contain any sequence after the first white space.

jowong4: You are correct about `read.id` and I'm now using `rfind`.

src/include/indexlr-workers.h
jowong4: I removed the assert and used an if statement instead because it should produce `NA  minimizers` instead of failing

src/include/indexlr-workers.h
```diff
@@ -286,6 +280,16 @@ MinimizeWorker::work()
 			} else {
```
sjackman:
```suggestion
			} else if (inputWorker.fasta) {
			    read.barcode = read.id;
			} else {
			    read.barcode = "NA";
			    …
			}
```

jowong4: Done

src/include/indexlr-workers.h
```diff
@@ -283,9 +277,19 @@ MinimizeWorker::work()
...
+				// For FASTA, use the sequence ID.
+				read.barcode = read.id;
 			} else {
-				// No barcode tag is present. For FASTA, use the sequence ID. For FASTQ, use NA.
-				read.barcode = inputWorker.fasta ? read.id : "NA";
+				// No barcode tag is present. Check for stLFR barcode within read.id.
+				read.barcode = "NA";
+				size_t sharpPos = read.id.find('#');
+				if (sharpPos != std::string::npos) {
+					size_t slashPos = read.id.rfind('/');
```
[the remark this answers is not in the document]
jowong4: Yeah I think I do need. I do not recall why I removed that line of code thought.

src/include/indexlr-workers.h
```diff
@@ -286,7 +286,7 @@ MinimizeWorker::work()
 				size_t sharpPos = read.id.find('#');
 				if (sharpPos != std::string::npos) {
 					size_t slashPos = read.id.rfind('/');
-					if (slashPos > sharpPos) {
+					if (slashPos != std::string::npos && slashPos > sharpPos) {
```
sjackman: It's okay if the read ID does not contain a slash. The `/1` and `/2` are optional. In that case, the barcode goes from the `#` to the end of the read ID.

jowong4: oh in that case, no explicit handling is required. `Number of characters to include in the substring (if the string is shorter, as many characters as possible are used).` frpm www.cplusplus.com/reference/string/string/substr/

jowong4: but if that is the case should I still use `rfind`? Isn't there a possibility the read identifier can contain "/". I don't think that is an illegal character.

sjackman: Anything's possible, but the read identifier does not usually contain `/` except for the `/1` and `/2` at the end. I'm okay with using `rfind`. You could also check whether the second-to-last character is a `/`.

jowong4: so if slash isn't necessary, there isn't a need to handle the case where `slashPos ==std::string::npos` because `string.substr` will retrieve as many characters as possible if the number of characters exceed length of string

PR #95, opened 2019-10-23
plotpaf.rmd: plot color and ideogram combined pdf

jowong4:
For the `x.paf.gz.q.pdf`, the smallest backbones are clumped together as other and isn't really helpful when we want to see how many backbone are in that one gray rectangle in the plot. 

This PR adds a new plot where I plot the backbones on the reference using an ideogram colour scheme. Then I overlay it with the largest backbones using the 25 colour scheme. This will achieve effectively the same result as `x.paf.gz.q.pdf` but instead of solid gray backbones for others, we get alternating gray in those sections

### DIFF
--- a/data/plotpaf.rmd
+++ b/data/plotpaf.rmd
@@ -175,8 +175,7 @@ pafq_ideogram <- pafq %>%
   group_by(Qname) %>%
   mutate(Colour = (Qindex + row_number()) %% 2) %>%
   ungroup() %>%
-  mutate(Colour = factor(Colour)) %>%
-  mutate(Colour = fct_recode(Colour, "Other1" = "0", "Other2" = "1"))
+  mutate(Colour = fct_recode(factor(Colour), "Other1" = "0", "Other2" = "1"))
 
 legend_breaks <- c(seq(0, 25, 1), c("Other1", "Other2"))
 
--- a/data/plotpaf.rmd
+++ b/data/plotpaf.rmd
@@ -180,7 +180,7 @@ pafq_ideogram <- pafq %>%
 
 legend_breaks <- c(seq(0, 25, 1), c("Other1", "Other2"))
 
-pafq_plot <- ggplot(pafq_ideogram) +
+ggplot(pafq_ideogram) +
   geom_rect(aes(
     xmin = Qstart, xmax = Qend,
     ymin = Qindex, ymax = Qindex + 1,
@@ -190,8 +190,7 @@ pafq_plot <- ggplot(pafq_ideogram) +
   scale_x_continuous(name = "Position", labels = unit_format(unit = "M", scale = 1e-6)) +
   scale_y_reverse(name = "Chromosome (Query)",
                   breaks = 0.5 + seq_len(nlevels(pafq$Qname)),
-                  labels = levels(pafq$Qname))
-pafq_plot  <- pafq_plot +
+                  labels = levels(pafq$Qname)) +
   geom_rect(data=pafq[!(pafq$Tname=="Other"),], aes(
     xmin = Qstart, xmax = Qend,
     ymin = Qindex, ymax = Qindex + 1,
@@ -204,7 +203,6 @@ pafq_plot  <- pafq_plot +
   theme_bw() +
   guides(fill = guide_legend(ncol = 1)) +
   labs(caption = input_paf)
-pafq_plot
 ```
 
 ```{r save-plot-q}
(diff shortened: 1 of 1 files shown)


sjackman:
Could you please post an example image?

jowong4:
Reference Coverage
[hg004.k40-w32.n100-5000.c2-x.physlr.overlap.n67.k3.mol.backbone.map.grch38.n10.qpos.paf.gz.q.pdf](https://github.com/bcgsc/physlr/files/3765131/hg004.k40-w32.n100-5000.c2-x.physlr.overlap.n67.k3.mol.backbone.map.grch38.n10.qpos.paf.gz.q.pdf)
Reference Coverage + ideogram
[hg004.k40-w32.n100-5000.c2-x.physlr.overlap.n67.k3.mol.backbone.map.grch38.n10.qpos.paf.gz.q+ideo.pdf](https://github.com/bcgsc/physlr/files/3765132/hg004.k40-w32.n100-5000.c2-x.physlr.overlap.n67.k3.mol.backbone.map.grch38.n10.qpos.paf.gz.q%2Bideo.pdf)

afshinfard:
That's nice.
One minor issue I see is that there are two shades of gray and then the contigs with these two colors are classified as target 26 and 27 while they're not just two. Am I right? I think its better to somehow cheat it and categorize them as others. With the current visualization the message we convey is that there two backbones that cover/map to multiple chromosomes.

jowong4:
> That's nice.
> One minor issue I see is that there are two shades of gray and then the contigs with these two colors are classified as target 26 and 27 while they're not just two. Am I right? I think its better to somehow cheat it and categorize them as others. With the current visualization the message we convey is that there two backbones that cover/map to multiple chromosomes.

I can try something like `Other` and `other`.

afshinfard:
> I can try something like `Other` and `other`.

Still confusing; but better than now for sure; and an easy fix.
Alternatively we can plot them all in a single grayscale and then plot lines as their boundaries with black or something. But I like the way you made it more. Isn't it possible to somehow group some those two legends in to one? not a big deal..,

sjackman:
Looks good! I think you can replace the existing `.q.pdf` plot with this one. It'd be nice to sort out the Target legend. I believe you could use a custom legend to set the text of both greys to `Other`, or if it's easier, you could label them `Other1` and `Other2`.

jowong4:
Ok, so it was not immediately apparent to me how I can combine colors into a box for the legend so I settled with `Other1` and `Other2`. This is what the picture looks like now.
Screenshot from 2019-10-24 10-13-13

data/plotpaf.rmd
```diff
@@ -166,21 +166,46 @@ pal["Other"] <- "#c0c0c0"
...
+  mutate(Colour = (Qindex + row_number()) %% 2 + 26) %>%
+  ungroup() %>%
+  mutate(Colour = factor(Colour))
+pafq_ideogram$Colour <- as.character(pafq_ideogram$Colour)
+pafq_ideogram$Colour[pafq_ideogram$Colour == "26"] <- "Other1"
+pafq_ideogram$Colour[pafq_ideogram$Colour == "27"] <- "Other2"
+pafq_ideogram$Colour <- as.factor(pafq_ideogram$Colour)
+pal <- setNames(alphabet_colours, pafq_lumped$Tname %>% setdiff("Other") %>% unique())
+pal["Other1"] <- "#808080"
+pal["Other2"] <- "#c0c0c0"
```
sjackman: Try to use `tidyverse` functions rather than `$` and `<-` on individual columns. Once you get in the habit, you'll find the code is more readable and easier to maintain. Use `mutate(Colour = …)` rather than `pafq_ideogram$Colour <-`. Look into the package `forcats` for modifying and reordering levels of a factor variable https://forcats.tidyverse.org. Read Hadley's book R for Data Science https://r4ds.had.co.nz. It's available online, and the lab has a copy. If the lab copy has grown legs and walked away, order a new copy.

jowong4: I used mutate to change the factor statements.

jowong4:
@sjackman  Do you have any other comments in this PR?

data/plotpaf.rmd
```diff
@@ -165,25 +165,44 @@ pafq_lumped <- paf_chained %>%
...
-	theme_bw() +
-	guides(fill = guide_legend(ncol = 1)) +
-	labs(caption = input_paf)
+pafq_ideogram <- pafq %>%
+  arrange(desc(Qlength), Qname, Qstart, Barcodes) %>%
+  group_by(Qname) %>%
+  mutate(Colour = (Qindex + row_number()) %% 2) %>%
+  ungroup() %>%
+  mutate(Colour = factor(Colour)) %>%
+  mutate(Colour = fct_recode(Colour, "Other1" = "0", "Other2" = "1"))
```
lcoombe: Can you merge these lines instead of being done separately?
```
mutate(Colour = fct_recode(factor(Colour), "Other1" = "0", "Other2" = "1"))
```

jowong4: done

PR #96, opened 2019-10-24
Report junctions

afshinfard:
aim: report junction-causing barcodes back to molecule stage so this stage do another round of molecule separation (more extensive this time) on only those vertices.
- a new method to report the junctions
- `determine_molecules()` now handles a second input file for junctions (as a graph with no edge) and then applies the requested set of algorithms only on those barcodes. It applies `bc` on all other barcodes not in junctions.
- and unrelated commit merged here but with no effect: `split_junctions_of_tree()` has a new input argument. when set to 0 it removes the junction instead of keeping the two largest edges. The default is set to 1 so no effect.
- I will add `Makefile` codes for reporting junctions and redoing mol-separation later.


afshinfard:
@lcoombe It's not what @jowong4 asked for and that one is already pushed into master. But I appreciate if you can review it if you had time.

physlr/physlr.py
```diff
@@ -1276,6 +1287,25 @@ def physlr_mst_troubleshoot(self):
...
+        gmst = Physlr.determine_pruned_mst(g)
+        print(int(timeit.default_timer() - t0), "Searching for junctions...", file=sys.stderr)
+        junctions = []
+        for component in nx.connected_components(gmst):
+            junctions += Physlr.detect_junctions_of_tree(
+                gmst.subgraph(component), Physlr.args.prune_junctions)
+        print(int(timeit.default_timer() - t0),
+              "Found", len(junctions), "junctions.", file=sys.stderr)
+        gjunctions = nx.Graph()
+        gjunctions.add_nodes_from(junctions)
```
lcoombe: Just wondering -- is it necessary to format this output as a graph? It looks like you are just adding nodes, but no edges, so couldn't you just use a TSV file of the junction nodes?

afshinfard: Right, I could. I found it easier at that time to do it that way as the method `write_tsv()` writes graphs in tsv format. I'll add a few lines of code to write the list as tsv.

physlr/physlr.py
```diff
@@ -537,9 +537,13 @@ def split_junctions_of_tree(prune_junctions, gin):
             if Physlr.args.verbose >= 3:
                 print("Junction:", junction, "Edges:", *edges, file=sys.stderr)
             # Keep the two incident edges with the largest weight, and remove the rest.
-            for u, v, _ in edges[2:-1]:
-                g.remove_edge(u, v)
-        return g
+            if keep_largest:
+                for u, v, _ in edges[2:-1]:
```
jowong4: shouldn't this code be
```suggestion
                for u, v, _ in edges[1:]:
```

physlr/physlr.py
```diff
@@ -1276,6 +1287,23 @@ def physlr_mst_troubleshoot(self):
...
+        gmst = Physlr.determine_pruned_mst(g)
+        print(int(timeit.default_timer() - t0), "Searching for junctions...", file=sys.stderr)
+        junctions = []
+        for component in nx.connected_components(gmst):
+            junctions += Physlr.detect_junctions_of_tree(
+                gmst.subgraph(component), Physlr.args.prune_junctions)
+        print(int(timeit.default_timer() - t0),
+              "Found", len(junctions), "junctions.", file=sys.stderr)
+        for junction in junctions:
+            print(junction, file=sys.stdout)
```
jowong4:
```suggestion
            print(junction)
```
I don't think the `file=sys.stdout` is necessary

physlr/physlr.py
```diff
@@ -1276,6 +1287,23 @@ def physlr_mst_troubleshoot(self):
...
+        print(int(timeit.default_timer() - t0), "Wrote the junctions.", file=sys.stderr)
```
jowong4:
```suggestion
        print(int(timeit.default_timer() - t0), "Wrote junctions to file.", file=sys.stderr)
```

physlr/physlr.py
```diff
@@ -1689,21 +1720,36 @@ def physlr_molecules(self):
...
                       "--separation-strategy: " + str(set(alg_list) - alg_white_list)
             sys.exit(exit_message)
+        junctions = []
+        if len(self.args.FILES) > 1:
+            with open(self.args.FILES[1]) as fin:
+                for line in fin:
+                    junctions.append(line.split()[0])
+            print(
+                int(timeit.default_timer() - t0),
+                "Separating junction resulting barcodes into molecules"
```
jowong4:
```suggestion
                "Separating junction (contributing / causing) barcodes into molecules"
```
(contributing / causing)  I think sounds better than resulting

physlr/physlr.py
```diff
@@ -1689,21 +1720,36 @@ def physlr_molecules(self):
             exit_message = "Error: physlr molecule: wrong input parameter(s) " + \
                       "--separation-strategy: " + str(set(alg_list) - alg_white_list)
             sys.exit(exit_message)
+        junctions = []
+        if len(self.args.FILES) > 1:
+            with open(self.args.FILES[1]) as fin:
```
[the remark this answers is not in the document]
jowong4: Yeah I think there is a bug here, gin doesn't get initialized if the if statement is true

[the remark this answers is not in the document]
afshinfard: Oh right, when I changed the code I forgot to consider this.

physlr/physlr.py
afshinfard: It keeps removes all the edges except the first two edges. so skipping 0 and 1 it should iterate from 2 to last. so [2:]. I'll fix it.

physlr/physlr.py
afshinfard: right

physlr/physlr.py
jowong4: It said `keep_largest` so I thought you only wanted the largest edge hence remove everything from 1 to last. So do you want to keep the two largest?

physlr/physlr.py
afshinfard: should I join them with a dash? like junction-causing?

lcoombe: Yeah I'd prefer `junction-causing`

physlr/physlr.py
lcoombe: We do specify `file=sys.stdout` throughout the physlr file though? I'd be OK keeping it for consistency.

physlr/physlr.py
afshinfard: the largest link between two of the branches, that includes two edges. yep.

physlr/physlr.py
afshinfard: Most of the time yes we do specif, and sometimes we don't. I'll keep it then.

afshinfard:
The commit msg is not corrected! Don't consider it. Is there an easy way to change it?

physlr/physlr.py
sjackman: Yes, that's a bug. It should be `[2:]`. What's the effect of that bug?

sjackman: Also 😠 😭

jowong4: hmm looks like I misunderstood the intention behind the code. That's all

jowong4:
> The commit msg is not corrected! Don't consider it. Is there an easy way to change it?

You can rebase -i and change the commit messages. You will have push --force at the end though

physlr/physlr.py
afshinfard: @sjackman no effects. This part of the code was not in the pipeline.

PR #97, opened 2019-10-25
Makefile: add rule to differentiate  %.qpos.paf.gz.pdf and %.paf.gz.pdf

jowong4:
So previously, there are no distinctions between plotting `%.qpos.paf.gz.pdf` and `%.paf.gz.pdf`. This has an unintended consequence with the ideogram output. The ideogram output uses the `gap.bed` to plot gaps in the reference. The problem is these positions are in chromosome positions plotting against minimizer positions. The result is the following:
Using gap.bed for %.paf.gz.pdf:
Screenshot from 2019-10-25 14-43-01

By generating a dummy file, I can resolve the problem
Using dummy.bed for %.paf.gz.pdf:
Screenshot from 2019-10-25 14-43-59

### DIFF
--- a/data/Makefile
+++ b/data/Makefile
@@ -1092,9 +1092,7 @@ mol_strategy=bc
 
 # Plot a PAF file.
 %.paf.gz.pdf: %.paf.gz
-	echo > dummy.bed; \
-	Rscript -e 'rmarkdown::render("$(physlr_path)/data/plotpaf.rmd", "html_document", "$*.paf.gz.html", knit_root_dir="$(PWD)", output_dir="$(PWD)", params=list(input_paf="$<", gap_bed="dummy.bed"))'; \
-	rm dummy.bed
+	Rscript -e 'rmarkdown::render("$(physlr_path)/data/plotpaf.rmd", "html_document", "$*.paf.gz.html", knit_root_dir="$(PWD)", output_dir="$(PWD)", params=list(input_paf="$<", gap_bed="dummy.bed"))'
 
 # Compare assembly metrics.
 %.quast.html: %.quast.tsv
--- a/data/Makefile
+++ b/data/Makefile
@@ -1086,9 +1086,15 @@ mol_strategy=bc
 %.bed.pdf: %.bed
 	Rscript -e 'rmarkdown::render("$(physlr_path)/data/plotbed.rmd", "html_document", "$*.bed.html", knit_root_dir="$(PWD)", output_dir="$(PWD)", params=list(input_bed="$<"))'
 
-# Plot a PAF file.
-%.paf.gz.pdf: %.paf.gz $(name)/$(ref).fa.gap.bed
+# Plot a qpos PAF file.
+%.qpos.paf.gz.pdf: %.qpos.paf.gz $(name)/$(ref).fa.gap.bed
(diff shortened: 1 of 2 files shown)


jowong4:
> Rather than having a dummy file, I'd rather either changing the Rmarkdown file to not plot the gaps on that particular ideogram, 

Problem is there is no Makefile distinction between the two different paf.gz.pdf, and if we add that distinction, then we can either use a dummy file or see below.

>or convert the gaps chromosome positions to minimizer positions.
>
> I think you'd be able to get at the latter approximately by using indexlr with `--pos` on the reference.
I don't think this would make sense. Plotting the reference coverage in minimizer position would have no gaps. It doesn't make sense to plot chromosome gaps in that case.


I'd be more inclined to add R code to allow empty string bed file argument.

lcoombe:
Correct me if I'm wrong, but I thought the issue was that we are generating multiple output PDFs from the one rule, and the gap BED file is fine for one, but not the other?

What are the output PDF files, and which of them are plotting the gaps? Because if we have separate rules, won't we just be running the same rule multiple times, generating the same PDF files twice, just one with gaps plotted and one without?

Are you saying that you don't want gaps plotted at all for the ideogram that you are showing above?

jowong4:
> Correct me if I'm wrong, but I thought the issue was that we are generating multiple output PDFs from the one rule, and the gap BED file is fine for one, but not the other?

Correct. To be more exact, only one should the use the BED file

> What are the output PDF files, and which of them are plotting the gaps? Because if we have separate rules, won't we just be running the same rule multiple times, generating the same PDF files twice, just one with gaps plotted and one without?

The `%.qpos.paf.gz.pdf` should have gaps and the %.paf.gz.pdf` shouldn't

> Are you saying that you don't want gaps plotted at all for the ideogram that you are showing above?

Correct.

lcoombe:
OK, but I still don't quite see why if they are separate PDF outputs generated by different R commands, why we wouldn't be able to just not plot the gaps on the graph in question?

I'm not sure I agree that the gaps are not useful. If the gap positions correspond to areas where there aren't backbones covering the position, I would think that is useful information.

data/Makefile
```diff
@@ -1086,10 +1086,14 @@ mol_strategy=bc
...
 
-# Plot a PAF file.
-%.paf.gz.pdf: %.paf.gz $(name)/$(ref).fa.gap.bed
+# Plot a qpos PAF file.
+%.qpos.paf.gz.pdf: %.qpos.paf.gz $(name)/$(ref).fa.gap.bed
 	Rscript -e 'rmarkdown::render("$(physlr_path)/data/plotpaf.rmd", "html_document", "$*.paf.gz.html", knit_root_dir="$(PWD)", output_dir="$(PWD)", params=list(input_paf="$<", gap_bed="$(name)/$(ref).fa.gap.bed"))'
 
+# Plot a PAF file.
+%.paf.gz.pdf: %.paf.gz
+	Rscript -e 'rmarkdown::render("$(physlr_path)/data/plotpaf.rmd", "html_document", "$*.paf.gz.html", knit_root_dir="$(PWD)", output_dir="$(PWD)", params=list(input_paf="$<", gap_bed="dummy.bed"))'
```
```suggestion
	Rscript -e 'rmarkdown::render("$(physlr_path)/data/plotpaf.rmd", "html_document", "$*.paf.gz.html", knit_root_dir="$(PWD)", output_dir="$(PWD)", params=list(input_paf="$<"))'
```

PR #98, opened 2019-11-04
Change pipeline

afshinfard:
Here I added some of my recent works for a better backbone detection (and miss-assembly correction) to the `Makefile`.

Those that affect the current pipeline:
- For the `backbone.tsv` we now remove all the remaining junctions with a 3rd largest branch of more than 200 nodes in the MST from the MST.

Other:
- We can run a 2nd round of molecule separation on only a list of junctions. For this to be the main pipleline we simply need to add `.mol2.` in the name of the file we want to make between `.mol.` and `.backbone.` so for example: 
`*.k3.mol.mol2.backbone.*`


afshinfard:
I am testing it at the monent.
BTW, any suggestions about the naming instead of `remol` is very welcome.

lcoombe:
What about `mol2` or `mol-r2` to indicate that it is another round of molecule separation? I'm wondering if it would be good to have it clear in the name that it is a second round? In case we ever want to do additional rounds?

afshinfard:
@lcoombe That's a good idea, I'll just change it. Also it'd be good if we could mention the strategy we use for the second round. I currently use bc+cos+sqcos but we can optimize that as well... but if I include them, the name would be too long then...

lcoombe:
Yes, good point about the strategy used being in the name. Perhaps we could just use the first letter of each separation strategy? Ie. `mol2-bcs`?
Do you think that the user will be setting the separation strategy or will we just set defaults that we've found work well?

afshinfard:
I feel like depending on the dataset we need some changes in the strategy but we might be able to automate it. for example for human and in the first round `bc` is kidna good for stflr but `bc+k3` is necessary for 10x. For the second round I feel it should be more general. We'll see.

data/Makefile
```diff
@@ -753,14 +756,30 @@ endif
...
+	$(time) $(python) $(bin)/physlr molecules -V$V -t$t --separation-strategy=bc+cos+sqcos $^ >$@
+
+# Redo molecule separation on a list of junction-causing barcodes.
+%.mol2-bkcs.tsv: %.tsv %.junctions.tsv
+	$(time) $(python) $(bin)/physlr molecules -V$V -t$t --separation-strategy=bc+k3+cos+sqcos $^ >$@
+
 # Determine the backbone graph from the overlap TSV.
 %.backbone.tsv: %.tsv
-	$(time) $(python) $(bin)/physlr backbone-graph --prune-branches=10 --prune-bridges=10 -s0 $< >$@
+	$(time) $(python) $(bin)/physlr backbone-graph --prune-branches=10 --prune-bridges=10 --prune-junctions=${prune_junction} -s0 $< >$@
```
```suggestion
	$(time) $(python) $(bin)/physlr backbone-graph --prune-branches=10 --prune-bridges=10 --prune-junctions=$(prune_junction) -s0 $< >$@
```

physlr/physlr.py
```diff
@@ -1839,6 +1839,8 @@ def index_minimizers_in_backbones(backbones, bxtomxs):
         mxtopos = {}
         for tid, path in enumerate(progress(backbones)):
             for pos, u in enumerate(path):
+                if u not in bxtomxs:
+                    u = u.rsplit("_", 1)[0]
```
sjackman: After the second round `AAAA_2` would become `AAAA_2_0` and `AAAA_2_1`? It seems a bit clumsy. I'd suggest either `AAAA_2` becomes `AAAA_2` and `AAAA_3` or possibly `AAAA_2` is removed and becomes `AAAA_3` and `AAAA_4`. I have a slight preference for the latter.
If we go with the current naming scheme of `AAAA_2_0`, these two `if` statements should be replaced with a `while` loop `while u not in bxtomxs and "_" in u`

afshinfard: Actually I just opened another branch to fix the things downstream so we can fix this. the problem with changing the number is that we need to double check and see how many molecules are there for this one that we are changing its name. For example if there we have `AAAA_0`, `AAAA_1`, and `AAAA_2` then for each of them in the second round we need to check the molecules list. If we deconvolute `AAAA_0` to `AAAA_3` and `AAAA_4` then we need to deconvolute `AAAA_1` to `AAAA_5` and `AAAA_6`. Is that fine? Or alternatively we can stick to the current naming and then go over the output once more to re-number all.

afshinfard: and my temporary fix is that we can duplicate the if statement in line 1842 and 1843 so it can go over the current output. I was about to open a PR for that. So let me know what's your suggestion.

afshinfard: Oh I did commit that duplicated if statement here... Sorry that was unintentional :D
The problem with while loop is that for stflr we have underscores in the naming, and we already know that we do at most 2 rounds of mol (for now, as it was a temporary fix) so if we change it we a while loop it will do unnecessary operations.

data/Makefile
```diff
@@ -839,6 +858,26 @@ min_path_size=200
...
+	$(time) $(python) $(bin)/physlr map-paf -V$V -n10 $^ | $(gzip) >$@
+
+# Map the reference to the (1:k3 / 2:bc+cos+sqcos) backbone graph and output PAF.
+%.overlap.n$n.k3.mol.mol2-bcs.backbone.map.$(ref).n10.paf.gz: %.overlap.n$n.k3.mol.mol2.backbone.path %.tsv $(name)/$(ref).k$k-w$w.physlr.tsv
+	$(time) $(python) $(bin)/physlr map-paf -V$V -n10 $^ | $(gzip) >$@
+
+# Map the reference to the (1:k3 / 2:bc+k3+cos+sqcos) backbone graph and output PAF.
+%.overlap.n$n.k3.mol.mol2-bkcs.backbone.map.$(ref).n10.paf.gz: %.overlap.n$n.k3.mol.mol2.backbone.path %.tsv $(name)/$(ref).k$k-w$w.physlr.tsv
+	$(time) $(python) $(bin)/physlr map-paf -V$V -n10 $^ | $(gzip) >$@
+
```
lcoombe: Is it possible to reduce the redundancy in rules here? Perhaps by using the wild card up to `backbone` and specifying the name for the `.tsv` file?

afshinfard: good point, umm, then I can change it to:
`%.map.$(ref).n10.paf.gz: %.path $(lr).k$k-w$w.n100-5000.c2-x.physlr.tsv $(name)/$(ref).k$k-w$w.physlr.tsv`
so the only problem is that then we will need multiple lines for other steps upstream (if it is `c2-x` or not) which is less redundant and we also need to define `Makefile` variables for n100-5000 too. Is that fine?

physlr/physlr.py
sjackman: You may want to consider changing the underscores in the stLFR barcodes names to say hyphens or some other character to avoid this source of confusion.

data/Makefile
lcoombe: Hmmm yes so it would only be the `c2-x` that could vary? Can you remind me what are the possibilities for that value? For our runs does that vary a lot?

afshinfard: I cannot recall well but I guess the other alternative is not to make `c2-x`, just skipping it... but we don't do that now so no variations in our current pipeline.

physlr/physlr.py
afshinfard: as it is in the name naming, wouldn't be more logical to change the symbol for molecule separation?
If we change the one in the stflr header we need to process more, if we change underscore in molecule separation then we need no more processing at all. What symbol do you think would be good? hyphen is already being used in the 10x headers. so `#` or `/` ?
and @jowong4 what are your suggestions? as you've already worked with stflr data and you know the symbols they use

data/Makefile
afshinfard: Done

physlr/physlr.py
jowong4: I think '~' might be fine. I don't recall seeing this symbol used in either of the files and seems like a decent substitute of `-` and `_`

afshinfard: anyway, is that fine if with postpone it to another PR and stick with this duplicated if statement?

data/Makefile
jowong4: @lcoombe  I don't think we have anything set up to change `c2-x`. It's always `c2-x`

lcoombe: Ok cool -- that's what I thought - Just wanted to check!

jowong4: I do however plan to add that into the code since there will be times where we need to change the C value, cough *spruce* cough

PR #99, opened 2019-11-06
Report junctions 2

afshinfard:
We can now report the junctions and their neighborhood of depth `--junction-depth`
The intention is that the junctions in the tree are not necessarily the undersplit vertex itself, and it might be a neighbor to that. This way we increase the chance of including the main undersplit vertex in the list of junctions to run the 2nd round of `molecules` on them.


physlr/physlr.py
```diff
@@ -1294,12 +1294,28 @@ def physlr_report_junctions_graph(self):
...
+            print(int(timeit.default_timer() - t0),
+                  "Exapnding junctions, depth:", self.args.junction_depth, ".", file=sys.stderr)
+            tree_junctions_expanded = []
+            for tree_junction in tree_junctions:
+                tree_junctions_expanded += list(nx.bfs_tree(
+                    gmst, source=tree_junction, depth_limit=self.args.junction_depth))
+            tree_junctions_expanded = list(set(tree_junctions_expanded))
+            junctions = [m for n in tree_junctions_expanded for m in g.neighbors(n)]
+            junctions += [n for n in tree_junctions_expanded]
+            junctions = list(set(junctions))
```
sjackman: I'd suggest avoiding converting from a list to a set back to a list? Modify lines 1312 and 131 to construct and update a set, then convert to a list in line 1314.

physlr/physlr.py
```diff
@@ -1294,12 +1294,28 @@ def physlr_report_junctions_graph(self):
...
         print(int(timeit.default_timer() - t0),
-              "Found", len(junctions), "junctions.", file=sys.stderr)
+              "Found", len(tree_junctions), "junctions.", file=sys.stderr)
+        junctions = []
+        if self.args.junction_depth > 0:
+            print(int(timeit.default_timer() - t0),
+                  "Exapnding junctions, depth:", self.args.junction_depth, ".", file=sys.stderr)
+            tree_junctions_expanded = []
+            for tree_junction in tree_junctions:
+                tree_junctions_expanded += list(nx.bfs_tree(
```
sjackman: Every `list(…)` is a warning sign that there may be a better way to do it. Avoid type conversions. In this case use `tree_junctions_expanded.extend(…)` instead of `tree_junctions_expanded += list(…)`.

physlr/physlr.py
```diff
@@ -1294,12 +1294,28 @@ def physlr_report_junctions_graph(self):
...
+                    gmst, source=tree_junction, depth_limit=self.args.junction_depth))
+            tree_junctions_expanded = list(set(tree_junctions_expanded))
```
sjackman: Avoid converting from a list to a set back to a list. `tree_junctions_expanded` should be a `set` while it's being constructed, then convert to a `list` at the end if you need a `list`, or maybe you don't need a `list` and it can be left as a `set`.

afshinfard:
@sjackman Changes are made, I appreciate if you review, thanks

physlr/physlr.py
```diff
@@ -686,7 +686,7 @@ def physlr_cut_chimera(g):
     @staticmethod
     def determine_pruned_mst(g):
         """Return the pruned maximum spanning tree of the graph."""
-        gmst = nx.maximum_spanning_tree(g, weight="n")
+        gmst = nx.maximum_spanning_tree(g, algorithm="kruskal", weight="n")
```
sjackman: Please add a comment why `kruskal`.

physlr/physlr.py
```diff
@@ -1294,15 +1294,33 @@ def physlr_report_junctions_graph(self):
...
+            tree_junctions += Physlr.detect_junctions_of_tree(
                 gmst.subgraph(component), Physlr.args.prune_junctions)
         print(int(timeit.default_timer() - t0),
-              "Found", len(junctions), "junctions.", file=sys.stderr)
+              "Found", len(tree_junctions), "junctions.", file=sys.stderr)
+        junctions = []
+        if self.args.junction_depth > 0:
+            print(int(timeit.default_timer() - t0),
+                  "Exapnding junctions, depth:", self.args.junction_depth, file=sys.stderr)
+            tree_junctions_expanded = set()
```
sjackman: Swap lines 1307 and 1308, move `tree_junctions_expanded = set()` inside the `if`.

PR #100, opened 2019-11-08
physlr: improve path_to_fasta

jowong4:
Improve path to fasta by inserting more Ns, and insert the equivalent number of bases as N when we have un-oriented pieces.

Change include:

- Default gap-size is 100 Ns (better results with Quast). Can be changed by user.

-  Print un-oriented contigs as Ns instead of removing them altogether (better results with Quast).

- Print unmapped contigs to avoid genome % drops (better results with Quast).

### DIFF
--- a/physlr/physlr.py
+++ b/physlr/physlr.py
@@ -2192,19 +2192,36 @@ class Physlr:
         num_scaffolds = 0
         num_contigs = 0
         num_bases = 0
+
+        unused_seqs = {name[0:-1] for path in paths for name in path if name[-1] == "."}
+
+        gaps = ("N" * self.args.gap_size)
+
         for path in progress(paths):
-            # Remove unoriented sequences.
-            path = [name for name in path if name[-1] != "."]
             if not path:
                 continue
 
-            seq = "NNNNNNNNNN".join(Physlr.get_oriented_sequence(seqs, name) for name in path)
+            seq = gaps.join(Physlr.get_oriented_sequence(seqs, name)
+                            if name[-1] != "." else ("N" * len(seqs[name[0:-1]]))
+                            for name in path)
+
             if len(seq) < self.args.min_length:
                 continue
             num_scaffolds += 1
             print(f">{str(num_scaffolds).zfill(7)} LN:i:{len(seq)} xn:i:{len(path)}\n{seq}")
             num_contigs += len(path)
             num_bases += len(seq)
+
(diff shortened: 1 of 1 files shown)


physlr/physlr.py
```diff
@@ -2437,6 +2454,9 @@ def parse_arguments():
             "--prune-junctions", action="store", dest="prune_junctions", type=int, default=0,
             help="split a backbone path when the alternative branch is longer than"
                  "prune-junctions [0]. set to 0 to skip.")
+        argparser.add_argument(
+            "--gap-size", action="store", dest="gap_size", type=int, default=100,
+            help="gap size used in scaffolding [0].")
```
```suggestion
            help="gap size used in scaffolding [100].")
```

physlr/physlr.py
```diff
@@ -2192,19 +2192,36 @@ def physlr_path_to_fasta(self):
...
             num_scaffolds += 1
             print(f">{str(num_scaffolds).zfill(7)} LN:i:{len(seq)} xn:i:{len(path)}\n{seq}")
             num_contigs += len(path)
             num_bases += len(seq)
+
+        for name in seqs:
+            if name in unused_seqs:
+                seq = seqs[name]
+                if len(seq) < self.args.min_length:
+                    continue
```
lcoombe: Do we print out all sequences -- even those that cannot be mapped to the backbone?

physlr/physlr.py
```diff
@@ -2192,19 +2192,36 @@ def physlr_path_to_fasta(self):
         num_scaffolds = 0
         num_contigs = 0
         num_bases = 0
+
+        unused_seqs = {name[0:-1] for path in paths for name in path if name[-1] == "."}
+
+        gaps = ("N" * self.args.gap_size)
```
```suggestion
        gaps = "N" * self.args.gap_size
```

physlr/physlr.py
jowong4: This was to address the drop in genome%

lcoombe: Yes that's the behavior we want - I just wanted to double check that we were in fact printing everything out (unoriented and unmapped)

jowong4: Oh I see what you mean

jowong4: The funny thing is I implemented it in the current way (printing unmapped) originally, but I changed to what you saw because at that time I couldn't remember why I implemented that way.

PR #102, opened 2019-11-08
ntHits and ntCard integration

jowong4:
This PR contains the initial steps to building a repetitive Bloom filter using kmers determined by ntHits with a repeat threshold cutoff calculated using the output of ntCard.

### DIFF
--- a/data/Makefile
+++ b/data/Makefile
@@ -87,6 +87,7 @@ endif
 .PHONY: f1chr4 f1chr2R f1 fishchr25 fish
 all: f1chr4 f1chr2R f1 fishchr25 fish
 
+
 ################################################################################
 # Fly
 # See https://support.10xgenomics.com/de-novo-assembly/datasets/2.1.0/fly
@@ -647,6 +648,15 @@ minsize=2000
 # Convert a .hist to a .histo file for GenomeScope.
 %.histo: %.hist
 	sed -n 's/^f//p' $< | tr '\t' ' ' >$@
+
+# Count k-mers using ntCard.
+%_k$(k).histogram.txt: %.fq.gz
+	ntcard -t $(t) -k $(k) -o $@ $<
+
+# Generate repetitive k-mers using ntHits.
+%_k$(k).rep: %.fq.gz %_k$(k).histogram.txt
+	$(eval repeat_threshold := $(shell $(python) $(bin)/physlr find-ntcard-mode $*_k$(k).histogram.txt ))
+	nthits -t $(t) -k $(k) -c $(repeat_threshold) -p $* $<;
 
 ################################################################################
 # Unicycler
--- a/data/Makefile
+++ b/data/Makefile
@@ -649,15 +649,19 @@ minsize=2000
 	sed -n 's/^f//p' $< | tr '\t' ' ' >$@
 
 # Count k-mers using ntCard.
-%_k$k.histogram.txt: %.fq.gz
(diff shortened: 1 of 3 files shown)


data/Makefile
```diff
@@ -641,6 +641,15 @@ minsize=2000
 %.histo: %.hist
 	sed -n 's/^f//p' $< | tr '\t' ' ' >$@
 
+# Count k-mers using ntCard.
+%_k$(k).histogram.txt: %.fq.gz
+	$(time) ntcard -t $(t) -k $(k) -o $@ $<
```
lcoombe: This is picky, but to keep it consistent -- we generally use `$k` for single-letter variables, and only the parentheses with multi-letter variables (ex. `$(time)`)

data/Makefile
```diff
@@ -641,6 +641,15 @@ minsize=2000
...
+
+# Generate repetitive k-mers using ntHits.
+%_k$(k).rep: %.fq.gz %_k$(k).histogram.txt
+	$(eval repeat_threshold := $(shell $(python) $(bin)/physlr find-ntcard-mode $*_k$(k).histogram.txt ))
+	$(time) nthits -t $(t) -k $(k) -c $(repeat_threshold) -p $* $<;
```
```suggestion
	$(time) nthits -t $(t) -k $(k) -c $(repeat_threshold) -p $* $<
```

data/Makefile
```diff
@@ -648,6 +648,15 @@ minsize=2000
 %.histo: %.hist
 	sed -n 's/^f//p' $< | tr '\t' ' ' >$@
 
+# Count k-mers using ntCard.
+%_k$(k).histogram.txt: %.fq.gz
```
```suggestion
%_k$k.histogram.txt: %.fq.gz
```

data/Makefile
```diff
@@ -648,6 +648,15 @@ minsize=2000
...
+	$(time) ntcard -t $t -k $k -o $@ $<
+
+# Generate repetitive k-mers using ntHits.
+%_k$(k).rep: %.fq.gz %_k$(k).histogram.txt
```
```suggestion
%_k$k.rep: %.fq.gz %_k$k.histogram.txt
```

data/Makefile
```diff
@@ -648,6 +648,18 @@ minsize=2000
...
 	sed -n 's/^f//p' $< | tr '\t' ' ' >$@
 
+# Count k-mers using ntCard.
+%_k$k.histogram.txt: %.fq.gz
+	$(time) ntcard -t $t -k $k -o $@ $<
+
+# Generate repetitive k-mers using ntHits.
+%_k$k.rep: %.fq.gz %_k$k.histogram.txt
+	$(eval mode := $(shell $(python) $(bin)/physlr find-ntcard-mode $*_k$k.histogram.txt ))
+	>&2 echo "The mode k-mer coverage is $(mode)."
```
lcoombe: We'll want `@echo` here so that it doesn't print out that line before it executes it. Do we need the `>&2`?

jowong4: For the second part I'm not sure. Im ok with piping it to stdout by default, but I figured some people might only use `2> x.log`

lcoombe:
```suggestion
	@>&2 echo "The mode k-mer coverage is $(mode)."
```
if you want to keep the stderr pipe

jowong4:
> Looks good to me! So now we have those runs, and we'll integrate into the default pipeline in future pull requests?

That's the idea. Eventually  `make xyz` will call those new rules that I added

PR #103, opened 2019-11-08
make Bloom filter to use with indexlr

jowong4:
Added code and rule to generate Bloom filter given ntHits output


data/Makefile
```diff
@@ -686,6 +686,11 @@ minsize=2000
 ################################################################################
 # Physlr
 
+# Generate repetitive Bloom filter to select minimizers
+bloom_filter_size=10000000000 #10G
```
lcoombe: I think we should put this with the other declared variables at the top of the Makefile

src/physlr-makebf.cc
```diff
@@ -0,0 +1,161 @@
...
+	std::cerr << progname << ": " << msg << "\nTry 'physlr-makebf --help' for more information.\n";
+}
+
+static void
+printUsage(const std::string& progname)
+{
+	std::cout << "Usage:  " << progname
+	          << "  -k K [-v] [-o FILE] FILE...\n\n"
+	             "  -k K       use K as k-mer size\n"
+	             "  -b B       Bloom Filter size in Bytes\n"
```
lcoombe:
```suggestion
	             "  -b B       Bloom filter size in Bytes\n"
```

src/physlr-makebf.cc
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+
+	BloomFilter bloomFilter(filterSize, hashNum, k);
+	if (verbose) {
+		std::cerr << "Made Bloom filter with:\n"
+		          << "kmer size                 = " << k << "\n"
+		          << "Bloom filter size         = " << filterSize << "\n"
+		          << std::endl;
+		std::cerr << "Inserting " << kmerVect.size() << " kmers to insert into bloom filter"
```
lcoombe:
```suggestion
		std::cerr << "Inserting " << kmerVect.size() << " kmers into bloom filter"
```

src/physlr-makebf.cc
```diff
@@ -0,0 +1,161 @@
...
+static void
+printUsage(const std::string& progname)
+{
+	std::cout << "Usage:  " << progname
+	          << "  -k K [-v] [-o FILE] FILE...\n\n"
+	             "  -k K       use K as k-mer size\n"
+	             "  -b B       Bloom Filter size in Bytes\n"
+	             "  -v         enable verbose output\n"
+	             "  -o FILE    write Bloom filter to FILE [required]\n"
+	             "  -t N       use N number of threads\n"
```
lcoombe:
```suggestion
	             "  -t N       use N number of threads [1]\n"
```
I believe the default behavior is to use 1 thread if the argument isn't specified?

src/physlr-makebf.cc
```diff
@@ -0,0 +1,161 @@
...
+	             "  --help     display this help and exit\n"
+	             "  FILE       space separated list of ntHits output files\n";
```
lcoombe: What file format for the ntHits output?

jowong4:
```
AATGGAATGGAATGGAACGAAATGTAATGTAATGGAATGG        143
AAACCAAACAGTAAACCAAATCCAATGAAAATAGAAGTCA        168
AATGGAATAGAAAGGAATGAACCCGAAGGGTATGGAATGG        147
ACAGCACAATTAACATAACACAACACAGCATAACACAACA        336
ACGTGAACATATACCCGTTTCGAATGAAGGCCAGCCAGTG        224
AAGGTGTGAACCCGGGAGGCGGAGCTTGCAGTGAGCCGAG        132
```
Example

lcoombe: You can do ntHits output in various formats thought, right? I was thining we might want to specify we want the TSV output.

PR #104, opened 2019-11-12
Deal with backbone with only un-oriented sequences

jowong4:
There are edges cases where a backbone consist of only un-oriented sequences. This PR contains code to deal with these exceptions, by taking the longest sequence within the backbone and assigning it a positive orientation

### DIFF
--- a/physlr/physlr.py
+++ b/physlr/physlr.py
@@ -2201,20 +2201,16 @@ class Physlr:
             if not path:
                 continue
 
-            if len(path) == 1:
-                if path[0][-1] == ".":
-                    path[0] = path[0][:-1] + "+"
-            else:
-                all_unoriented = True
-                longest_seq = 0
-                for idx, name in enumerate(path):
-                    if name[-1] != ".":
-                        all_unoriented = False
-                        break
-                    if len(name) > len(path[longest_seq]):
-                        longest_seq = idx
-                if all_unoriented:
-                    path[longest_seq] = path[longest_seq][:-1] + "+"
+            all_unoriented = True
+            longest_seq = 0
+            for idx, name in enumerate(path):
+                if name[-1] != ".":
+                    all_unoriented = False
+                    break
+                if len(name) > len(path[longest_seq]):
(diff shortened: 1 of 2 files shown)


physlr/physlr.py
```diff
@@ -2195,14 +2195,23 @@ def physlr_path_to_fasta(self):
...
+            for idx, name in enumerate(path):
+                if name[-1] != ".":
+                    all_unoriented = False
+                    break
+                if len(name) > len(path[longest_seq]):
+                    longest_seq = idx
+            if all_unoriented:
+                path[longest_seq] = path[longest_seq][:-1] + "+"
+
             seq = gaps.join(Physlr.get_oriented_sequence(seqs, name)
```
lcoombe: Wouldn't we get the same sequences outputted if we check if all the sequences are unoriented, and just move on to the next path in those cases? So just don't place anything here if all are unoriented, then the code below will take care of placing that sequence. 
Just wondering if that's cleaner.

jowong4: Your solution is more correct. It just becomes harder for me to track down the backbone. That's why I did this. Thought I suppose in the end, it doesn't really matter.

jowong4: addressed

physlr/physlr.py
```diff
@@ -2195,11 +2195,20 @@ def physlr_path_to_fasta(self):
...
+            all_unoriented = True
+            longest_seq = 0
+            for idx, name in enumerate(path):
+                if name[-1] != ".":
+                    all_unoriented = False
+                    break
+                if len(name) > len(path[longest_seq]):
+                    longest_seq = idx
+            if all_unoriented:
+                path = None
```
lcoombe: What about something like:
```
all_unoriented = all([name[-1] == "." for name in path])
if all_unoriented:
    continue
```
 I don't think you need the length tracking anymore

jowong4: That is correct. Didn't know there there was an `all` function

PR #107, opened 2019-11-14
0.2.0 Pipeline

jowong4:
This is WIP, but I want your input on variable names and various makefile matters


data/Makefile
```diff
@@ -100,46 +147,46 @@ all: f1chr4 f1chr2R f1 fishchr25 fish
 f1chr4: \
 	f1chr4.k32-w32.n100-1000.physlr.mxperbx.tsv.pdf \
 	f1chr4.k32-w32.n100-1000.physlr.depth.tsv.pdf \
-	f1chr4.k32-w32.n100-1000.c2-x.physlr.overlap.n$n.mol.backbone.path.$(ref).molecule.bed.pdf \
-	f1chr4.k32-w32.n100-1000.c2-x.physlr.overlap.n$n.mol.backbone.map.$(ref).n10.paf.gz.pdf \
-	f1chr4.k32-w32.n100-1000.c2-x.physlr.overlap.n$n.mol.backbone.map.$(draft).n10.sort.best.bed.pdf \
-	f1chr4.k32-w32.n100-1000.c2-x.physlr.overlap.n$n.mol.backbone.map.$(draft).n10.sort.best.bed.path.$(ref).minidot.pdf \
-	f1chr4.k32-w32.n100-1000.c2-x.physlr.overlap.n$n.mol.backbone.label.gv.pdf \
-	f1chr4.k32-w32.n100-1000.c2-x.physlr.overlap.n$n.mol.backbone.map.$(ref).n10.ann.gv.squish.pdf
+	f1chr4.k32-w32.n100-1000.c2-x.physlr.overlap.n$m.mol.backbone.path.$(ref).molecule.bed.pdf \
```
afshinfard: issues of being fast like Johnathan :D you forgot to change the `n` harcoded in the naming.
`f1chr4.k32-w32.n100-1000.c2-x.physlr.overlap.m$m.mol.backbone.path.$(ref).molecule.bed.pdf`

data/Makefile
```diff
@@ -90,6 +103,47 @@ endif
...
+	@echo ""
+	@echo "    Options:"
+	@echo "	draft           draft name [draft]. File must have .fa extension"
+	@echo "	ref           ref name [draft]. File must have .fa extension. If provided, will generate Quast metrics and backbone metrics relative to reference"
+	@echo "	lr           read name [reads]. File must have .fq.gz extension"
+	@echo "	t		number of threads used [16]"
+	@echo "	minimizer_overlap		% of edges to remove in the overlap graph [90]. Available options are stLFR, 10X, or [0-100]."
+	@echo "	k		size of k-mers [32]."
+	@echo "	w		window size to choose k-mers [32]."
+	@echo "	min_component_size		minimum number of barcodes within a backbone to output [50]."
```
```suggestion
	@echo "	min_component_size		minimum number of barcodes in a backbone [50]."
```

data/Makefile
```diff
@@ -90,6 +103,47 @@ endif
...
+	@echo "	minimum_barcode_multiplicity		minimum number of reads a barcode can have [10]."
```
lcoombe: Is it min/max number of reads, or number of minimizers for a barcode?

data/Makefile
```diff
@@ -90,6 +103,47 @@ endif
...
+	@echo "	maximum_barcode_multiplicity		maximum number of reads a barcode can have [5000]."
+	@echo "	minimizer_multiplicity		maximum number of edges an edges can contribute to [100]."
+	@echo "                             Alternatively, you can specify x, which uses the Q3 + IQR of the minimizer multiplicity distribution"
```
lcoombe: Hmm I don't really understand this -- do you specify a parameter 'x'? Or do you specify 'x' for the above parameter?

data/Makefile
```diff
@@ -90,6 +103,47 @@ endif
...
+	@echo "	bloom_filter_size		size of bloom filter [10000000000]."
+	@echo ""
+	@echo "Example: To run physlr with myDraft.fa, myReads.fq.gz, and stLFR data, run:"
+	@echo "	./make physical draft=myDraft reads=myReads minimizer_overlap=stLFR"
```
```suggestion
	@echo "	./make physical-map draft=myDraft reads=myReads minimizer_overlap=stLFR"
```

data/Makefile
```diff
@@ -948,12 +987,27 @@ g=$(shell awk '{x += $$2} END{print x}' $(name)/$(ref).fa.fai)
...
+	$(time) $(python) $(bin)/physlr filter-overlap --minimizer-overlap 10 $< >$@
+
+# Filter 20% of lowest weighted edges using python.
+%.m20.tsv: %.tsv
+	$(time) $(python) $(bin)/physlr filter-overlap --minimizer-overlap 20 $< >$@
+
+# Filter 50% of lowest weighted edges using python.
+%.m50.tsv: %.tsv
+	$(time) $(python) $(bin)/physlr filter-overlap --minimizer-overlap 50 $< >$@
+
```
lcoombe: Just wondering why we need these explicit rules when we have the rule using the variable `m`?

data/Makefile
```diff
@@ -90,6 +103,47 @@ endif
...
+	@echo "	minimum_barcode_multiplicity		minimum number of reads a barcode can have [10]."
+	@echo "	maximum_barcode_multiplicity		maximum number of reads a barcode can have [5000]."
+	@echo "	minimizer_multiplicity		maximum number of edges an edges can contribute to [100]."
+	@echo "                             Alternatively, you can specify x, which uses the Q3 + IQR of the minimizer multiplicity distribution"
+	@echo "	bloom_filter_size		size of bloom filter [10000000000]."
+	@echo ""
+	@echo "Example: To run physlr with myDraft.fa, myReads.fq.gz, and stLFR data, run:"
+	@echo "	./make physical draft=myDraft reads=myReads minimizer_overlap=stLFR"
+	@echo "To ensure that the pipeline runs correctly, make sure that the following tools are in your PATH: ntCard, ntHits, and Quast"
+
```
lcoombe: Are we going to let the user decide on the molecule separation strategy? Or just keep it distributed always?

physlr/physlr.py
```diff
@@ -1239,6 +1239,46 @@ def physlr_overlap(self):
         self.write_tsv(g, sys.stdout)
         print(int(timeit.default_timer() - t0), "Wrote the graph", file=sys.stderr)
 
+    def physlr_filter_overlap(self):
+        "Read a Physlr overlap graph and filter edges."
+
+        edge_weight = []
+
+        at_edges = False
+        overlap_input = open(self.args.FILES[0], "r")
```
```suggestion
        with open(self.args.FILES[0], "r") as overlap_input:
```
I prefer that style of opening/working with files

data/Makefile
```diff
@@ -90,6 +103,47 @@ endif
...
+ifndef ref
+physical-map: \
+	$(lr).k$k-w$w.n$(n1)-$(n2).c2-$(minimizer_multiplicity).physlr.overlap.m$m.mol.mol2-bcs.backbone.path
+else
+physical-map: \
+	$(lr).k$k-w$w.n$(n1)-$(n2).c2-$(minimizer_multiplicity).physlr.overlap.m$m.mol.mol2-bcs.backbone.map.$(ref).n10.paf.gz.pdf
+endif
+# Help
+help:
+	@echo "Usage: ./make [COMMAND] [OPTION=VALUE]..."
```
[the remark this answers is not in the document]
jowong4: that is probably going to be the last thing in the PR or in a separate PR for that matter. The file change interface doesn't work very well with moving files and changing code at the same time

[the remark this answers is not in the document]
lcoombe: Ok if it is in a separate PR we should change the name of this PR then. Because I thought that basically when you merged this, it should be in release-ready form.

data/Makefile
jowong4: Well I added the `mol_strategy` in the help section but I don't expect them to change it

physlr/physlr.py
jowong4: the code here is in a separate PR to keep things separate

lcoombe: hmm Ok probably it would have been good to not commit the code to this PR then -- that's a bit confusing

data/Makefile
```diff
@@ -10,6 +10,12 @@ t=16
 # Size of Bloom filter
 bloom_filter_size=10000000000 #10GB
 
+# Molecule separation stratergy
+mol_strategy=distributed
+
+# maximum number of edes an minimizer can contribute to
```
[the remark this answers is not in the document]
jowong4: They are the same thing, but if the comment there is to explain what changing the `minimzier_multiplicity` does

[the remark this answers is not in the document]
lcoombe: Personally, I find that description confusing.

data/Makefile
jowong4: hmm i forgot how the code does that. I think it's minimizers now that you mentioned it

data/Makefile
jowong4: these explicit rules are legacies of the past. I can remove them though.

data/Makefile
```diff
@@ -90,6 +103,47 @@ endif
...
+	@echo "	scaffold	run physlr and scaffold an existing assembly with a physical map"
+	@echo "	help            display this help page"
+	@echo "	version         display the software version"
+	@echo ""
+	@echo "    Options:"
+	@echo "	draft           draft name [draft]. File must have .fa extension"
+	@echo "	ref           ref name [draft]. File must have .fa extension. If provided, will generate Quast metrics and backbone metrics relative to reference"
+	@echo "	lr           read name [reads]. File must have .fq.gz extension"
+	@echo "	t		number of threads used [16]"
+	@echo "	minimizer_overlap		% of edges to remove in the overlap graph [90]. Available options are stLFR, 10X, or [0-100]."
```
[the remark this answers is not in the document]
jowong4: done

data/Makefile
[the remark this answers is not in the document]
jowong4: Comment has been change to `Maximum minimizer multiplicity`

data/Makefile
jowong4: No I meant, you can specify x to that parameter. I've changed the format of that line. Hopefully that clarifies the intention

data/Makefile
```diff
@@ -1,5 +1,5 @@
 # Size of a k-mer.
-k=32
+k=40
```
lcoombe: Is this the k we are using for all human? Just wondering why we changed the default

data/Makefile
```diff
@@ -36,15 +39,26 @@ name=$(ref)
...
+else ifeq ($(minimizer_overlap), 10X)
+m=92.5
+else
+m=$(minimizer_overlap)
+endif
 
 # Minimum/Maximum barcode multiplicity
 n1=100
 n2=5000
+ifdef minimum_barcode_multiplicity
```
lcoombe: Why don't we just use the full variable name? Just thinking then we wouldn't need this code section

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
 .DELETE_ON_ERROR:
 .SECONDARY:
 
-.PHONY: f1chr4 f1chr2R f1 fishchr25 fish
+.PHONY: f1chr4 f1chr2R f1 fishchr25 fish physical-map scaffolds
 all: f1chr4 f1chr2R f1 fishchr25 fish
 
+$(lr).physlr.physical-map: $(lr).k$k-w$w.n$(n1)-$(n2).c2-$(minimizer_multiplicity).physlr.overlap.m$m.mol.mol2-bcs.backbone.path
```
```suggestion
$(lr).physlr.physical-map.path: $(lr).k$k-w$w.n$(n1)-$(n2).c2-$(minimizer_multiplicity).physlr.overlap.m$m.mol.mol2-bcs.backbone.path
```

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+	ln -sf $< $@
+
+$(lr).physlr.physical-map.$(ref).n10.paf.gz: $(lr).k$k-w$w.n$(n1)-$(n2).c2-$(minimizer_multiplicity).physlr.overlap.m$m.mol.mol2-bcs.backbone.map.$(ref).n10.paf.gz.pdf
+	ln -sf $< $@
+
+$(draft).physlr.fa: $(lr).k$k-w$w.n$(n1)-$(n2).c2-$(minimizer_multiplicity).physlr.overlap.m$m.mol.mol2-bcs.backbone.map.$(draft).n10.sort.best.bed.path.fa
```
lcoombe: Should there be a soft link command here?

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
 .SECONDARY:
 
-.PHONY: f1chr4 f1chr2R f1 fishchr25 fish
+.PHONY: f1chr4 f1chr2R f1 fishchr25 fish physical-map scaffolds
 all: f1chr4 f1chr2R f1 fishchr25 fish
 
+$(lr).physlr.physical-map: $(lr).k$k-w$w.n$(n1)-$(n2).c2-$(minimizer_multiplicity).physlr.overlap.m$m.mol.mol2-bcs.backbone.path
+	ln -sf $< $@
+
+$(lr).physlr.physical-map.$(ref).n10.paf.gz: $(lr).k$k-w$w.n$(n1)-$(n2).c2-$(minimizer_multiplicity).physlr.overlap.m$m.mol.mol2-bcs.backbone.map.$(ref).n10.paf.gz.pdf
```
```suggestion
$(lr).physlr.physical-map.$(ref).n10.paf.gz.pdf: $(lr).k$k-w$w.n$(n1)-$(n2).c2-$(minimizer_multiplicity).physlr.overlap.m$m.mol.mol2-bcs.backbone.map.$(ref).n10.paf.gz.pdf
```

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+scaffolds: \
+	physical-map \
+	$(draft).physlr.quast.tsv
+endif
+
+physical-map:
+
+ifndef ref
+physical-map: \
+	$(lr).physlr.physical-map
```
```suggestion
	$(lr).physlr.physical-map.path
```

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+else
+physical-map: \
+	$(lr).physlr.physical-map \
```
```suggestion
	$(lr).physlr.physical-map.path \
```

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+	@echo ""
+	@echo "    Options:"
+	@echo "	draft               draft name [draft]. File must have .fa extension"
+	@echo "	ref                 reference name [ref]. File must have .fa extension. If provided, will generate Quast metrics and backbone visualization relative to reference."
+	@echo "	lr                  read name [reads]. File must have .fq.gz extension"
+	@echo "	t		    number of threads used [16]"
+	@echo "	k		    size of k-mers [40]."
+	@echo "	w		    window size to choose minimizers [32]."
+	@echo "	minimizer_overlap		% of edges to remove in the overlap graph [90]. Available options are stLFR, 10X, or [0-100]."
+	@echo "					Default values for stLFR and 10X were obtained by finding a value that work well across multiple datasets"
```
lcoombe: I think I was unclear here -- I mean mention what the default values are for stLFR and 10x

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+	@echo "					Increasing the parameter would generate more conservative results."
```
```suggestion
	@echo "					Increasing the parameter generates more conservative results."
```

data/Makefile
lcoombe: Do you think this change shouldn't be made? Wasn't sure if you resolved accidentally

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+	@echo "	min_component_size		minimum number of barcodes within a backbone to output [50]."
+	@echo "	minimum_barcode_multiplicity	minimum number of minimizers a barcode can have [10]."
```
```suggestion
	@echo "	minimum_barcode_multiplicity	minimum number of minimizers per barcode [10]."
```

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+	@echo "	maximum_barcode_multiplicity	maximum number of minimizers a barcode can have [5000]."
```
```suggestion
	@echo "	maximum_barcode_multiplicity	maximum number of minimizers per barcode [5000]."
```

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+	@echo "	minimizer_multiplicity		maximum number of edges a minimizer can contribute to [100]. Specify x to use the Q3 + IQR of the minimizer multiplicity distribution instead."
```
```suggestion
	@echo "	minimizer_multiplicity		maximum minimizer multiplicity [100]. Specify x to use the Q3 + IQR of the minimizer multiplicity distribution instead."
```

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+	@echo "	mol_strategy			molecule separation strategy [distributed]. Available options are bc, bc+k3, distributed."
+	@echo "					bc is the least conservative and is only suitable for datasets with low barcode multiplicity"
```
lcoombe: I think we should at least mention what the abbreviations stand for

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+	@echo "					bc+k3 is most conservative and requires more time"
+	@echo "					distributed is a modified version of bc+k3 that is faster but less conservative"
+	@echo "	bloom_filter_size		size of bloom filter [10000000000] (10G)."
+	@echo ""
+	@echo "Example: To get a physial map using physlr with stLFR reads myReads.fq.gz, run:"
```
```suggestion
	@echo "Example: To generate a physical map using physlr with stLFR reads myReads.fq.gz, run:"
```

data/Makefile
```diff
@@ -87,70 +101,139 @@ endif
...
+	@echo "	./make physical-map lr=myReads minimizer_overlap=stLFR"
+	@echo "Example: To scaffold a draft assembling using physlr with 10X reads myReads.fq.gz and draft.fa, run:"
```
```suggestion
	@echo "Example: To scaffold a draft assembly using physlr with 10X reads myReads.fq.gz and draft.fa, run:"
```

data/Makefile
```diff
@@ -423,11 +506,11 @@ hg004.quast.tsv: \
 		hg004.abyss.quast.tsv \
 		hg004.supernova.quast.tsv \
 		hg004.supernova.scaftigs.quast.tsv \
-		hg004.k40-w32.n100-5000.c2-x.physlr.overlap.n62.k3.mol.backbone.map.hg004.abyss.n10.sort.best.bed.path.quast.tsv \
-		hg004.k40-w32.n100-5000.c2-x.physlr.overlap.n62.k3.mol.backbone.map-mkt.hg004.abyss.n10.sort.best.bed.path.quast.tsv \
-		hg004.k40-w32.n100-5000.c2-x.physlr.overlap.n62.k3.mol.backbone.map.hg004.supernova.scaftigs.n10.sort.best.bed.path.quast.tsv \
-		hg004.k40-w32.n100-5000.c2-x.physlr.overlap.n62.k3.mol.backbone.map-mkt.hg004.supernova.scaftigs.n10.sort.best.bed.path.quast.tsv \
-		hg004.k40-w32.n100-5000.c2-x.physlr.overlap.n62.k3.mol.backbone.map.hg004.supernova.n10.sort.best.bed.path.quast.tsv
+		hg004.k40-w32.n100-5000.c2-$(minimizer_multiplicity).physlr.overlap.n62.k3.mol.backbone.map.hg004.abyss.n10.sort.best.bed.path.quast.tsv \
```
lcoombe: We're changing `n` to `m`, right?

data/Makefile
lcoombe: I'm OK with removing them - doesn't seem necessary to me

data/Makefile
jowong4: the old rules are not changed for result compatibility. n= uses absolute weight filtering where as m uses % of edges

data/Makefile
jowong4: I don't think the path suffix is necessary. Path isn't a standard suffix to begin with so I think labeling physical path will work too

lcoombe: Hmm I disagree. That is a suffix we use for ABySS, and I feel like we need to indicate the file format generally somehow

data/Makefile
lcoombe: Ok makes sense!

data/Makefile
jowong4: Yes we are using k40 for all human and higher k is more accurate.

data/Makefile
jowong4: Im pretty sure that was the problem I had with m that I mentioned to you on slack, it doesn't assign the value properly.

data/Makefile
jowong4: I do not think the path is necessary since path isn't a standard suffix anyways.

data/Makefile
jowong4: Yes, there should be because the mistake is dependency here. It should be 'paf.gz' instead of `paf.gz.pdf`. The reasoning is `plotpaf.rmd` generates multiple pdf files so soft linking one of them is not very useful. The intention was to softlink it to a new prefix and make the pdfs with the new prefix

data/Makefile
jowong4: see above

data/Makefile
jowong4: Well I was going to add in the readme that the output is with x suffix and they are in y format

data/Makefile
jowong4: Actually you are right, in this case, it doesnt matter since you dont have to do text processing

PR #108, opened 2019-11-15
Permit float values in paf file

jowong4:
Post chaining paf files can have float mapq and the code as it is will break


sjackman:
mapq are integers. PAF files should not have a float mapq. That's a bug in the program that created that PAF file.

PR #110, opened 2019-11-15
Change Makefile to physlr-make

jowong4:
Renamed the Makefile to physlr-make similar to how arc does it

@lcoombe, is the credit order ok?


sjackman:
I suggest renaming `data/Makefile` to `bin/physlr-make`, that is, put it in the `bin` directory. The `data` subdirectory could include a symlink `Makefile` to `../bin/physlr-make` for convenience.

lcoombe:
I agree with Shaun -- I think it makes more sense to have the newly named makefile in the `bin` directory.

PR #112, opened 2019-11-20
Parallelize Overlap stage of physlr-overlap

jowong4:
Parallelize Physlr-overlap by computing all the edges of a given barcode. The drawback is you will revisit the same minimizer index in `minimizerToBarcode`. New code improves overlap speed by ~10 folds at 48 threads for the overlap stage

Note: The ordering of output is not deterministic. Making it deterministic will remove pretty much any time improvement.


lcoombe:
Did you track the improvement in peak memory as well?

jowong4:
`hg004.k40-w32.n100-5000.c2-x.physlr.tsv`
|   | Time (hh:mm:ss) | Memory (GB)|
| --- | --- | --- |
|Before| 2:11:04 |  200|
|After| 0:45:03 | 351 |

@lcoombe

lcoombe:
> `hg004.k40-w32.n100-5000.c2-x.physlr.tsv`
> | Results | Time (hh:mm:ss) | Memory (GB)|
> | --- | --- | --- |
> |Before| 2:11:04 | 200|
> |After| 0:45:03 | 351 |
> 
> @lcoombe

Huh so the peak memory usage is worse?

jowong4:
I accidentally mixed up the memory cost in the table. It has been updated

src/Makefile
```diff
@@ -61,3 +61,7 @@ check-physlr-index-stLFR-10X: all
 check-physlr-makebf: all
 	./physlr-makebf -t16 -k8 -b1000 -o tiny.bf data/tiny_k8.rep
 	./physlr-indexlr -t16 -k8 -w1 -r tiny.bf data/stLFR.tiny.fq | diff -q - data/stLFR.tiny.filtered.physlr.tsv.good
+
+check-physlr-overlap: all
+	./physlr-overlap -t1 -n1 data/tiny.overlap.input.tsv | sort | diff -q - data/tiny.physlr.overlap.n1.sorted.tsv
+	./physlr-overlap -t1 -n2 data/tiny.overlap.input.tsv | sort | diff -q - data/tiny.physlr.overlap.n2.sorted.tsv
```
lcoombe: Perhaps if you are parallelizing, it would be good to have a test that tests using multiple threads?

src/physlr-overlap.cc
```diff
@@ -191,94 +192,102 @@ main(int argc, char* argv[])
...
+	// print out vertexes + counts
+	for (const auto& itr : barcodes) {
+		bufferString.clear();
+		bufferString += itr.first;
+		bufferString += "\t";
+		bufferString += std::to_string(barcodeToMinimizer[itr.second].size());
+		bufferString += "\n";
+		std::cout << bufferString;
+	}
+	std::cout << "\nU\tV\tn" << std::endl;
```
lcoombe: Are we still calling this `n` in the code, or should we commit to the `m` change?

jowong4:
Added time metrics for newest version

src/physlr-overlap.cc
jowong4: I think if the file uses `m` in the name then it is good enough. If change it here, I will also have to chage the python read graph code

src/Makefile
jowong4: This PR has been rebased on the unit test PR. I will change it there

src/physlr-overlap.cc
lcoombe: @schutzekatze - What do you think?? I think it's a bit confusing to have 2 names for the parameter... A user couldn't really understand the columns in that file if they wanted to?

I'd prefer just make the full change to `m` to avoid confusing future maintainers of the code/users, but that's just my opinion

jowong4: If we do decide to change it, it will be on another PR. Do you have any other comments. If not i will merge this to master

jowong4:
For posterity 
Spruce

|  Results |  Memory (GB)|
| --- | --- |
|Before (C20)| ~1,300 |
|After (C100)| 600|

PR #114, opened 2019-11-22
plotpaf: permit float map values

jowong4:
`ploftpaf.rmd` breaks if the map values are not integers


jowong4:
@lcoombe  merge this for now and start an issue to remind ourselves? I don't know if the bug will be difficult to solve

jowong4:
see #117 for resolution

PR #118, opened 2019-11-25
Makefile: add install rule

jowong4:
Add install rule to Makefile

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,6 +1,6 @@
 CXXFLAGS=-g -O3 -Wall -Werror -Wno-error=unknown-pragmas -std=c++11 -pthread -I./include -I./vendor -fopenmp
 LDLIBS=-lz
-PREFIX = $(HOME)/.local
+PREFIX=$(HOME)/.local
 
 .PHONY: all clean check install lint clang-format clang-tidy fix check-physlr-index-fasta check-physlr-index-pos-fasta check-physlr-index-fastq check-physlr-filter-barcodes check-physlr-filter-bxmx-first check-physlr-filter-bxmx-second check-physlr-index-stLFR check-physlr-index-stLFR-10X check-physlr-makebf check-physlr-overlap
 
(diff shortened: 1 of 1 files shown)


lcoombe:
Just curious -- the binaries will be installed to that specified location, but I guess the users will likely never use them there because the Makefile specifically points to the `src` directory in the repo?

jowong4:
Yeah, that is currently the case. I'm guessing this is intended for the more advanced users who want to use/configure parts of physlr so they add the install path to their `PATH` variable. Can call the binaries individually.

src/Makefile
```diff
@@ -1,5 +1,8 @@
 CXXFLAGS=-g -O3 -Wall -Werror -Wno-error=unknown-pragmas -std=c++11 -pthread -I./include -I./vendor -fopenmp
 LDLIBS=-lz
+PREFIX = $(HOME)/.local
```
sjackman:
```suggestion
PREFIX=$(HOME)/.local
```

src/Makefile
```diff
@@ -65,3 +68,7 @@ check-physlr-makebf: all
 check-physlr-overlap: all
 	./physlr-overlap -t4 -n1 data/tiny.overlap.input.tsv | sort | diff -q - data/tiny.physlr.overlap.n1.sorted.tsv
 	./physlr-overlap -t4 -n2 data/tiny.overlap.input.tsv | sort | diff -q - data/tiny.physlr.overlap.n2.sorted.tsv
+
+install: physlr-indexlr physlr-filter-barcodes physlr-overlap physlr-filter-bxmx physlr-makebf
+	mkdir -p $(DESTDIR)$(PREFIX)/bin
```
sjackman:
```suggestion
	install -d $(DESTDIR)$(PREFIX)/bin
```

src/Makefile
```diff
@@ -65,3 +68,7 @@ check-physlr-makebf: all
...
+	for f in $^; do cp $$f $(DESTDIR)$(PREFIX)/bin; done
```
sjackman:
```suggestion
	install $^ $(DESTDIR)$(PREFIX)/bin
```

src/Makefile
jowong4: i used `install $^ -t $(DESTDIR)$(PREFIX)/bin` instead. Is that alright?

src/Makefile
jowong4: see above

src/Makefile
jowong4: Nevermind, the above command doesn't seem to make the directory if it doesn't exist

PR #109, opened 2019-11-15
Populate README

jowong4:
Populated README with usage and dependency information

PR #122, opened 2019-11-26
Makefile protocol variable

jowong4:
per Shaun's request in #109, I'm separating `minimizer_overlap` into `minimizer_overlap` and `protocol` where the options for protocol are "stlfr" or "10x" and specifying it would override `minimizer_overlap`.


bin/physlr-make
```diff
@@ -41,14 +41,18 @@ endif
...
 m=85
-else ifeq ($(minimizer_overlap), 10X)
+else ifeq ($(protocol), 10x)
 m=92.5
+endif
 else
+ifdef minimizer_overlap
 m=$(minimizer_overlap)
+else
+m=90
```
lcoombe: Can you remind me why we need both `m` and `minimizer_overlap`? It seems to me just having `protocol` and `m` would suffice?

bin/physlr-make
```diff
@@ -169,10 +173,12 @@ help:
 	@echo " 			1: periodic messages."
 	@echo " 			2: progress bar."
 	@echo " 			3: verbose messages."
-	@echo "	minimizer_overlap		% of edges to remove in the overlap graph [90]. Available options are stLFR, 10X, or [0-100]."
-	@echo "					stLFR: 85."
-	@echo "					10X: 92.5."
-	@echo "					Default values for stLFR and 10X were obtained by finding a value that worked well across multiple data sets."
+	@echo "	minimizer_overlap, protocol	% of edges to remove in the overlap graph [90]. Available options are [0-100]."
```
lcoombe: I think we should separate these to different lines

bin/physlr-make
```diff
@@ -185,9 +191,11 @@ help:
 	@echo "	bloom_filter_size		size of bloom filter [10000000000] (10G)."
 	@echo ""
 	@echo "Example: To generate a physical map using physlr with stLFR reads myReads.fq.gz, run:"
-	@echo "	./physlr-make physical-map lr=myReads minimizer_overlap=stLFR"
+	@echo "	./physlr-make physical-map lr=myReads protocol=stLFR"
+	@echo "Example: To generate a physical map using physlr with linked reads myReads.fq.gz and a customize minimizer_overlap threshold, run:"
```
```suggestion
	@echo "Example: To generate a physical map using physlr with linked reads myReads.fq.gz and a specified minimizer_overlap threshold, run:"
```

bin/physlr-make
jowong4: the idea was the user would be more informed if he uses the long form of `m` which is `minimizer_overlap` but internally we use` m`, I could change all the `m` to `minimizer_overlap` or vice verse

lcoombe: Right.. I suppose the name isn't as descriptive now that we are removing a % of edges vs. using an absolute minimizer threshold. I'd be OK with changing it all to `m` -- the help page is there for the user to understand what it controls.

bin/physlr-make
```diff
@@ -169,11 +168,13 @@ help:
...
 	@echo " 			2: progress bar."
 	@echo " 			3: verbose messages."
-	@echo "	minimizer_overlap		% of edges to remove in the overlap graph [90]. Available options are stLFR, 10X, or [0-100]."
-	@echo "					stLFR: 85."
-	@echo "					10X: 92.5."
-	@echo "					Default values for stLFR and 10X were obtained by finding a value that worked well across multiple data sets."
+	@echo "	m 	% of edges to remove in the overlap graph [90]. Available options are [0-100]."
+	@echo "	protocol		Available options are stlfr and 10x. Specify the protocol option to use the default m value for the respective linked read protocol."
+	@echo "					stlfr: 85."
+	@echo "					10x: 92.5."
```
```suggestion
	@echo "					10x: m=92.5."
```

bin/physlr-make
```diff
@@ -169,11 +168,13 @@ help:
...
+	@echo "					Default values for stlfr and 10x were obtained by finding a value that worked well across multiple data sets."
 	@echo "					Increasing this parameter would generate more conservative results."
+	@echo "					NOTE: protocol option overrides m option."
```
```suggestion
	@echo "					NOTE: protocol option overrides 'm'."
```

bin/physlr-make
```diff
@@ -169,11 +168,13 @@ help:
 	@echo " 			1: periodic messages."
 	@echo " 			2: progress bar."
 	@echo " 			3: verbose messages."
-	@echo "	minimizer_overlap		% of edges to remove in the overlap graph [90]. Available options are stLFR, 10X, or [0-100]."
-	@echo "					stLFR: 85."
-	@echo "					10X: 92.5."
-	@echo "					Default values for stLFR and 10X were obtained by finding a value that worked well across multiple data sets."
+	@echo "	m 	% of edges to remove in the overlap graph [90]. Available options are [0-100]."
+	@echo "	protocol		Available options are stlfr and 10x. Specify the protocol option to use the default m value for the respective linked read protocol."
```
lcoombe: Are we going to use all lowercase or mixed upper-case/lowercase? I know Shaun suggested using `stlfr` all lower-case, and it could make sense to make that change in this PR.
Either way, we need to be consistent.

bin/physlr-make
```diff
@@ -185,9 +186,11 @@ help:
 	@echo "	bloom_filter_size		size of bloom filter [10000000000] (10G)."
 	@echo ""
 	@echo "Example: To generate a physical map using physlr with stLFR reads myReads.fq.gz, run:"
-	@echo "	./physlr-make physical-map lr=myReads minimizer_overlap=stLFR"
+	@echo "	./physlr-make physical-map lr=myReads protocol=stLFR"
```
lcoombe: Let's update these examples to reflect Shaun's naming suggestion while we're here to avoid using capitals. 
Ie. `myreads.fq.gz` instead of `myReads.fq.gz`

bin/physlr-make
```diff
@@ -185,9 +186,11 @@ help:
...
+	@echo "Example: To generate a physical map using physlr with linked reads myReads.fq.gz and a specified m threshold, run:"
+	@echo "	./physlr-make physical-map lr=myReads m=86"
 	@echo "Example: To scaffold a draft assembly using physlr with 10X reads myReads.fq.gz and draft.fa, run:"
-	@echo "	./physlr-make scaffolds lr=myReads minimizer_overlap=10X draft=draft"
+	@echo "	./physlr-make scaffolds lr=myReads protocol=10X draft=draft"
```
```suggestion
	@echo "	./physlr-make scaffolds lr=myReads protocol=10x draft=draft"
```
I think you changed to use lower-case `x`?

bin/physlr-make
```diff
@@ -169,11 +168,13 @@ help:
 	@echo " 			1: periodic messages."
 	@echo " 			2: progress bar."
 	@echo " 			3: verbose messages."
-	@echo "	minimizer_overlap		% of edges to remove in the overlap graph [90]. Available options are stLFR, 10X, or [0-100]."
-	@echo "					stLFR: 85."
-	@echo "					10X: 92.5."
-	@echo "					Default values for stLFR and 10X were obtained by finding a value that worked well across multiple data sets."
+	@echo "	m 	% of edges to remove in the overlap graph [90]. Available options are [0-100]."
+	@echo "	protocol		Available options are stlfr and 10x. Specify the protocol option to use the default m value for the respective linked read protocol."
+	@echo "					stlfr: 85."
```
```suggestion
	@echo "					stlfr: m=85."
```

bin/physlr-make
jowong4: oh i forgot to change the examples

PR #123, opened 2019-11-27
physlr-make: use k-mer repetitive threshold instead of 100

jowong4:
As discussed in the abyss 10 meeting, we should use a more empirical value for minimizer multiplicity than the arbitrary 100 that I set. As k-mer multiplicity is related to minimizer multiplcity, we are going to use that.

### DIFF
--- a/bin/physlr-make
+++ b/bin/physlr-make
@@ -17,9 +17,6 @@ bloom_filter_size=10000000000 #10GB
 # Molecule separation stratergy
 mol_strategy=distributed
 
-# Maximum minimizer multiplicity
-minimizer_multiplicity=100
-
 # Path to the Physlr project.
 physlr_path=$(shell dirname $$(dirname $(realpath $(MAKEFILE_LIST))))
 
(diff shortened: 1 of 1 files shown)


bin/physlr-make
```diff
@@ -177,7 +177,7 @@ help:
 	@echo "	min_component_size		minimum number of barcodes in a backbone [50]."
 	@echo "	minimum_barcode_multiplicity	minimum number of minimizers per barcode [10]."
 	@echo "	maximum_barcode_multiplicity	maximum number of minimizers per barcode [5000]."
-	@echo "	minimizer_multiplicity		maximum minimizer multiplicity [100]. Specify x to use the Q3 + IQR of the minimizer multiplicity distribution instead."
+	@echo "	minimizer_multiplicity		Available option is x. Specify x to use the Q3 + IQR of the minimizer multiplicity distribution instead of the repetitive k-mer threshold."
```
lcoombe: So here are you saying that you can only specify `x`? That's the only valid option? 
If that's the case, it would make more sense to make it more boolean (1 for on, 0 for off, say)

jowong4: I've decided to allow the user to specify their own minimizer_multiplicity

bin/physlr-make
```diff
@@ -177,7 +174,9 @@ help:
 	@echo "	min_component_size		minimum number of barcodes in a backbone [50]."
 	@echo "	minimum_barcode_multiplicity	minimum number of minimizers per barcode [10]."
 	@echo "	maximum_barcode_multiplicity	maximum number of minimizers per barcode [5000]."
-	@echo "	minimizer_multiplicity		maximum minimizer multiplicity [100]. Specify x to use the Q3 + IQR of the minimizer multiplicity distribution instead."
+	@echo "	minimizer_multiplicity		maximum minimizer multiplicity [repetitive k-mer threshold based on ntCard]."
+	@echo "					Specify x to use the Q3 + IQR of the minimizer multiplicity distribution instead."
```
```suggestion
	@echo "					Specify x to use the Q3 + IQR of the minimizer multiplicity distribution."
```

jowong4:
As `k-mer multiplicity` is a run time variable, it is quite difficult to get it work as part of the rule, so we will revert that back to `x` as the default name. User should not have to change that variable and if we want to test other `minimizer_multiplicity` values, we can do so by manually changing

PR #125, opened 2019-11-30
Integration test

jowong4:
To prime #124 , I have added an integration test for all physlr modules from filtering overlap file to generating the path file. In the future, this test will hopefully catch anything that we break that we didn't catch.


bin/physlr-make
```diff
@@ -909,9 +918,12 @@ endif
 %.mol2-bkcs.tsv: %.tsv %.junctions.tsv
 	$(time) $(python) $(bin)/physlr molecules -V$V -t$t --separation-strategy=bc+k3+cos+sqcos $^ >$@
 
+prune_branches=10
```
lcoombe: So we needed to add in these checks to make the integration test work? I think we should have these at the top of the file with the other parameter declarations

azure-pipelines.yml
```diff
@@ -27,3 +27,5 @@ jobs:
     displayName: Run clang-format and clang-tidy
   - script: make -C src check
     displayName: Test C++ executables
+  - script: make -C data check python_executable=python m=0 prune_branches=0 prune_bridges=0
```
lcoombe:
```suggestion
  - script: make -C data test python_executable=python m=0 prune_branches=0 prune_bridges=0
```
Just a suggestion to make it more clear this is a test

bin/physlr-make
```diff
@@ -192,6 +196,11 @@ help:
 version:
 	@echo "physlr v0.2.0"
 
+################################################################################
+# Test m=0 prune_branches=0 prune_bridges=0
+test: \
```
lcoombe: Does the test not work with m>0, or doing pruning? The downside of this is that we won't be testing any of the code working with those parameters

jowong4: I could change `m` to something else and change the test a bit. No clue about `prune_branches` though. Could try 1 and see if that would still give me an aouput

lcoombe: Ideally the small test would be as representative as possible of a typical physlr run

bin/physlr-make
jowong4: I dont think the users should touch these variables just as they dont touch `minimizer_multiplicity`

bin/physlr-make
jowong4: Well the intention was to test if the code would run properly, so I think as long as it does executes the code that would have ran in a typical physlr run, it will be fine. So I will change the test set to accommodate m greater than 0 but the other parameter is little harder to work around.

bin/physlr-make
lcoombe: Well then maybe put them by the test code instead? Just seems strange to have them here by the pruning code if we wouldn't actually touch them in any other situation other than the test

jowong4: Those values are default values of `prune_branches`. Previously they were hardcoded

lcoombe: Yes I know -- We kept them hardcoded because we didn't want to change them in the normal physlr case. If we do _only_ want to change them in the test situation, then I think it makes more sense to put them by that relevant code.

PR #129, opened 2019-12-18
Physlr molecules C++ port

jowong4:
This PR contains the biconnected component C++ port of physlr molecules.


src/physlr-molecules.cc
```diff
@@ -0,0 +1,412 @@
...
+		exit(EXIT_SUCCESS);
+	} else if (infiles.empty()) {
+		printErrorMsg(progname, "missing file operand");
+		failed = true;
+	}
+	if (t > 1) {
+		std::cerr << "physlr-molecules does not support multithreading currently." << std::endl;
+		t = 1;
+	}
+	if (separationStrategy != "bc") {
```
lcoombe: I'm not sure it makes sense to have the other separation strategies in the help page right now if they aren't actually supported in the code? We can update everything to be consistent as more are added in.

src/physlr-molecules.cc
```diff
@@ -0,0 +1,412 @@
...
+	if (help != 0) {
+		printVersion();
+		printUsage(progname);
+		exit(EXIT_SUCCESS);
+	} else if (infiles.empty()) {
+		printErrorMsg(progname, "missing file operand");
+		failed = true;
+	}
+	if (t > 1) {
+		std::cerr << "physlr-molecules does not support multithreading currently." << std::endl;
```
lcoombe:
```suggestion
		std::cerr << "physlr-molecules does not support multithreading currently. Reverting to 1 thread" << std::endl;
```
Is this something that you are going to actively work on soon? If it's only something we'll do further in the future we could remove that option for now.

src/physlr-molecules.cc
lcoombe: Also doesn't look like this variable is ever used anywhere right now?

src/physlr-molecules.cc
```diff
@@ -0,0 +1,412 @@
...
+	}
+
+	std::cerr << "Finished molecule separation ";
+#if _OPENMP
+	std::cerr << "in sec: " << omp_get_wtime() - sTime << std::endl;
+	sTime = omp_get_wtime();
+#endif
+	std::cerr << "Memory usage: " << double(memory_usage()) / double(1048576) << "GB" << std::endl;
+
+	std::cerr << "Generating new graph" << std::endl;
```
lcoombe:
```suggestion
	std::cerr << "Generating molecule overlap graph" << std::endl;
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,412 @@
...
+	std::cout << "U\tn" << std::endl;
+	std::string node1, node2;
+	int weight;
+	auto vertexItRange = boost::vertices(g);
+	for (auto vertexIt = vertexItRange.first; vertexIt != vertexItRange.second; ++vertexIt) {
+		node1 = g[*vertexIt].name;
+		weight = g[*vertexIt].weight;
+		std::cout << node1 << "\t" << weight << "\n";
+	}
+	std::cout << "\nU\tV\tn" << std::endl;
```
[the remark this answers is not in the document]
jowong4: Will be all done in #124

src/physlr-molecules.cc
```diff
@@ -0,0 +1,412 @@
...
+	}
+#if _OPENMP
+	std::cerr << "in sec: " << omp_get_wtime() - sTime << std::endl;
+	sTime = omp_get_wtime();
+#endif
+	std::cerr << "Memory usage: " << double(memory_usage()) / double(1048576) << "GB" << std::endl;
+}
+
+void
+componentsToNewGraph(
```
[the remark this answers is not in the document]
jowong4: This is for when you want to run other molecule separation on the results of the previous biconnected component output. It is possible to print the final without this but I think this is better.

src/physlr-molecules.cc
```diff
@@ -0,0 +1,412 @@
...
+    const graph_t& oldG,
```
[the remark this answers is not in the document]
jowong4: `oldG` is the input graph. You need to `oldG`'s vertex and edge information to generate `newG`

src/physlr-molecules.cc
jowong4: Hmm I can remove the other mol strategy from the help, but Im going to keep the `separationStrategy` variable

src/physlr-molecules.cc
jowong4: Something for the future. `Physlr-overlap` has the same thing when it was single thread only

src/physlr-molecules.cc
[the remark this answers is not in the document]
lcoombe: So `newG` is not a molecule overlap graph??

src/physlr-molecules.cc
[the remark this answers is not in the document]
lcoombe: Yup the code is fine, but a comment in the code clarifying the purpose is that I was suggesting.

src/physlr-molecules.cc
lcoombe: I just think it strange to have that option if we aren't going to be working on it in the next month or so (which you may be, I just didn't know).
It isn't hard to add an option when the feature is added - better to do that in my opinion then have a useless option.

src/physlr-molecules.cc
jowong4: done

src/physlr-molecules.cc
jowong4: updated

src/physlr-molecules.cc
```diff
@@ -0,0 +1,412 @@
...
+static void
+printUsage(const std::string& progname)
+{
+	std::cout << "Usage:  " << progname
+	          << "  [-s SEPARATION-STRATEGY] [-v -t T] FILE...\n\n"
+	             "  -v         enable verbose output\n"
+	             "  -s --separation-strategy   \n"
+	             "  -t N       use N number of threads [1]\n"
+	             "  --help     display this help and exit\n"
+	             "  SEPARATION-STRATEGY      + separated list of molecule separation strategies "
```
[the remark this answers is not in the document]
jowong4: done

src/physlr-molecules.cc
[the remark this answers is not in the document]
jowong4: oh comment as in in the code?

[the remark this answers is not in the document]
lcoombe: Yup :)

src/physlr-molecules.cc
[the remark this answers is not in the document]
jowong4: `newG` is a molecule overlap graph. I'm just saying `oldG` isn't necessarily a barcode graph.

src/physlr-molecules.cc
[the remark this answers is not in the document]
jowong4: done

src/physlr-molecules.cc
jowong4: done

src/physlr-molecules.cc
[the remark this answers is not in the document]
lcoombe: I just can't get past the 'old graph'/'new graph' terms -- they aren't very descriptive.
It's fine for the variables I guess, but for the function name and the comment in the code.
What about like 'Generate a molecule overlap graph based on molecule separation. Input graph is barcode overlap graph or molecule overlap graph from previous round of molecule separation'? Or something like that..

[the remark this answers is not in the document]
jowong4: I don't mind changing it but I feel it is akin to `infile` and `outfile`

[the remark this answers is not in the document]
lcoombe: Yeah sure for the variables I get it -- I'm talking about the comment

src/physlr-molecules.cc
```diff
@@ -0,0 +1,398 @@
...
+
+static void
+printUsage(const std::string& progname)
+{
+	std::cout << "Usage:  " << progname
+	          << "  [-s SEPARATION-STRATEGY] [-v] FILE...\n\n"
+	             "  -v         enable verbose output\n"
+	             "  -s --separation-strategy   \n"
+	             "  --help     display this help and exit\n"
+	             "  SEPARATION-STRATEGY      `+` separated list of molecule separation strategies "
```
lcoombe: I think I suggested before that the separation strategy help information would be best to be by the `-s`?

jowong4: I must have missed it if you did. It's done now.

src/physlr-molecules.cc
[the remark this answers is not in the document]
jowong4: Is it better now?

src/physlr-molecules.cc
```diff
@@ -0,0 +1,398 @@
...
+biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
+{
+	// Find biconnected components
+	boost::property_map<graph_t, edgeComponent_t>::type component =
+	    boost::get(edgeComponent, subgraph);
+	boost::biconnected_components(subgraph, component);
+
+	std::vector<vertex_t> artPointsVec;
+	articulation_points(subgraph, std::back_inserter(artPointsVec));
+	std::unordered_set<vertex_t> art_points(artPointsVec.begin(), artPointsVec.end());
```
afshinfard: So we call `boost::biconnected_components(subgraph, component)` and then we call `articulation_points(subgraph, std::back_inserter(artPointsVec))`. I am afraid that the bottleneck in computing biconnected components is calculating articulation points and we're calling them twice so it doubles the running time. If it's true, isn't it possible to call only `boost::biconnected_components(subgraph, component)` and do the job based on that?

afshinfard: I just figured out that by calling `boost::biconnected_components(subgraph, component)` the components are overlapping by their articulation points, and you need two calls so you remove articulation points from communities to make it consistent with our python implementation.
Although using only `biconnected_components()` make it different from what we had in python but I always wanted to switch to this mode where we do not through out the articluation points themseves.... May produce better results... we can test it later...

afshinfard: Not related to this PR, but if it makes sense for you can you please check if muting the `articulation_points` call would reduce the run-time?

jowong4: I had intended to fix this but I forgot. The two functions can be combined

afshinfard:
Thanks for the change. Looks good to me too!

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+
+#if _OPENMP
+#include <omp.h>
+#endif
+
+#define PROGRAM "physlr-molecules"
+#define PACKAGE_NAME "physlr"
+#define GIT_REVISION "pre-autotools"
+
+static int
```
sjackman: Prefer `size_t` or `unsigned` for unsigned integers.

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+struct edgeProperties
+{
+	int weight = 0;
+};
+
+struct edgeComponent_t
+{
+	enum
+	{
+		num = 555
```
sjackman: Use `std::numeric_limits<int>::max()` or `INT_MAX`.

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+
+void
+printGraph(const graph_t& g)
+{
+	std::cout << "U\tn" << std::endl;
+	std::string node1, node2;
+	int weight;
+	auto vertexItRange = boost::vertices(g);
+	for (auto vertexIt = vertexItRange.first; vertexIt != vertexItRange.second; ++vertexIt) {
+		node1 = g[*vertexIt].name;
```
sjackman:
```suggestion
		auto& node1 = g[*vertexIt].name;
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+		weight = g[*vertexIt].weight;
```
sjackman:
```suggestion
		auto& weight = g[*vertexIt].weight;
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+	             "  SEPARATION-STRATEGY      `+` separated list of molecule separation strategies "
+	             "[bc]\n"
+	             "  --help     display this help and exit\n";
+}
+
+void
+printGraph(const graph_t& g)
+{
+	std::cout << "U\tn" << std::endl;
+	std::string node1, node2;
```
sjackman:
```suggestion
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+	int weight;
```
sjackman:
```suggestion
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+		node1 = g[*vertexIt].name;
+		weight = g[*vertexIt].weight;
+		std::cout << node1 << "\t" << weight << "\n";
+	}
+	std::cout << "\nU\tV\tn" << std::endl;
+	auto edgeItRange = boost::edges(g);
+	for (auto edgeIt = edgeItRange.first; edgeIt != edgeItRange.second; ++edgeIt) {
+		weight = g[*edgeIt].weight;
+		node1 = g[boost::source(*edgeIt, g)].name;
+		node2 = g[boost::target(*edgeIt, g)].name;
```
sjackman: Ditto.

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+					sTime = omp_get_wtime();
+#endif
+				}
+				continue;
+			}
+
+			std::string node1, node2;
+			int weight;
+			std::istringstream ss(line);
+			ss >> node1;
```
sjackman: What happens if this extraction fails? Nearly every extraction needs to be wrapped in an `if`.
```suggestion
			if (ss >> node1) {
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+			if (!atEdges) {
+				ss >> weight;
```
sjackman: Ditto.

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+				u = boost::add_vertex(g);
+				g[u].name = node1;
+				g[u].weight = weight;
+				g[u].indexOriginal = u;
+				barcodeToIndex[node1] = u;
+				indexToBarcode[u] = node1;
+			} else {
+				ss >> node2 >> weight;
```
sjackman: Ditto.

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+				continue;
+			}
+
+			std::string node1, node2;
+			int weight;
+			std::istringstream ss(line);
+			ss >> node1;
+			if (!atEdges) {
+				ss >> weight;
+				u = boost::add_vertex(g);
```
sjackman: Move the declaration to the point of initialization whenever possibe.
```suggestion
				auto u = boost::add_vertex(g);
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+	barcodeToIndex_t barcodeToIndex;
+	indexToBarcode_t indexToBarcode;
+	for (auto& infile : infiles) {
+		infile == "-" ? "/dev/stdin" : infile;
+		vertex_t u;
+		edge_t E;
+		std::ifstream infileStream(infile);
+		std::string line;
+		bool atEdges = false;
+		while (std::getline(infileStream, line)) {
```
sjackman:
```suggestion
		for (std::string line; std::getline(infileStream, line;)) {
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+
+	barcodeToIndex_t barcodeToIndex;
+	indexToBarcode_t indexToBarcode;
+	for (auto& infile : infiles) {
+		infile == "-" ? "/dev/stdin" : infile;
+		vertex_t u;
+		edge_t E;
+		std::ifstream infileStream(infile);
+		std::string line;
+		bool atEdges = false;
```
sjackman: Remove all or as many as possible of these variable declarations.

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+componentsToNewGraph(
+    const graph_t& inG,
+    graph_t& molSepG,
+    vecVertexToComponent_t& vecVertexToComponent)
+{
+	barcodeToIndex_t molSepGBarcodeToIndex;
+#if _OPENMP
+	double sTime = omp_get_wtime();
+#endif
+	for (size_t i = 0; i < vecVertexToComponent.size(); i++) {
```
sjackman:
```suggestion
	for (size_t i = 0; i < vecVertexToComponent.size(); ++i) {
```
Always use preincrement `++x` and avoid postincrement `x++`.

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+			molSepG[u].weight = inG[i].weight;
+			molSepG[u].indexOriginal = u;
+			molSepGBarcodeToIndex[molSepG[u].name] = u;
+		}
+	}
+
+	auto edgeItRange = boost::edges(inG);
+	for (auto edgeIt = edgeItRange.first; edgeIt != edgeItRange.second; ++edgeIt) {
+		vertex_t u, v;
+		u = inG[boost::source(*edgeIt, inG)].indexOriginal;
```
sjackman:
```suggestion
		auto u = inG[boost::source(*edgeIt, inG)].indexOriginal;
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+			molSepG[u].name = inG[i].name + "_" + std::to_string(j);
+			molSepG[u].weight = inG[i].weight;
+			molSepG[u].indexOriginal = u;
+			molSepGBarcodeToIndex[molSepG[u].name] = u;
+		}
+	}
+
+	auto edgeItRange = boost::edges(inG);
+	for (auto edgeIt = edgeItRange.first; edgeIt != edgeItRange.second; ++edgeIt) {
+		vertex_t u, v;
```
sjackman:
```suggestion
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+		u = inG[boost::source(*edgeIt, inG)].indexOriginal;
+		v = inG[boost::target(*edgeIt, inG)].indexOriginal;
```
sjackman:
```suggestion
		auto v = inG[boost::target(*edgeIt, inG)].indexOriginal;
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+
+	// Remove components with size less than 1
+	for (auto&& vertexSet : componentToVertexSet) {
+		if (vertexSet.size() <= 1) {
+			continue;
+		}
+		for (auto&& vertex : vertexSet) {
+			vertexToComponent[vertex] = moleculeNum;
+		}
+		moleculeNum++;
```
sjackman:
```suggestion
		++moleculeNum;
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+	static int help = 0;
+	std::string separationStrategy = "bc";
+	bool verbose = false;
+	bool failed = false;
+	static const struct option longopts[] = {
+		{ "help", no_argument, &help, 1 },
+		{ "separation-strategy", required_argument, nullptr, 's' },
+		{ nullptr, 0, nullptr, 0 }
+	};
+	while ((c = getopt_long(argc, argv, "s:v", longopts, &optindex)) != -1) {
```
sjackman:
```suggestion
	for (int c; (c = getopt_long(argc, argv, "s:v", longopts, &optindex)) != -1;) {
```

src/physlr-molecules.cc
```diff
@@ -0,0 +1,401 @@
...
+		moleculeNum++;
+	}
+}
+
+int
+main(int argc, char* argv[])
+{
+
+	auto progname = "physlr-molecules";
+	int c;
```
sjackman:
```suggestion
```

src/physlr-molecules.cc
jowong4: done

src/physlr-molecules.cc
jowong4: Done.

src/physlr-molecules.cc
jowong4: Done.

src/physlr-molecules.cc
jowong4: I changed that for loop into two for loops.

src/physlr-molecules.cc
jowong4: Done.

src/physlr-molecules.cc
jowong4: Done.

src/physlr-molecules.cc
jowong4: Done.

src/physlr-molecules.cc
jowong4: Done.

src/physlr-molecules.cc
jowong4: Done and exit when conditional is false.

src/physlr-molecules.cc
jowong4: Done and exit when conditional is false.

src/physlr-molecules.cc
jowong4: Done and exit when conditional is false.

sjackman:
Very exciting work! `physlr molecules` was a major bottle neck in the Physlr pipeline.

PR #131, opened 2019-12-18
Use minimizers of each molecule rather than the barcode to map sequences

jowong4:
Fly 

|Assembly|NG50|NGA50|# misassemblies|Genome fraction (%)|
| --- | --- | --- | --- |--- |
|Original|20,504,419|839,829|467|89.025|
|Edge intersection|18,548,285|720,198|329|77.852|
|Split minimizer based on graph|20,495,800|890,313|450|89.017|
|Split minimizer based on graph + orient with more pieces|20,495,800|968,954|473|88.988|

### DIFF
--- a/physlr/physlr.py
+++ b/physlr/physlr.py
@@ -1269,6 +1269,41 @@ class Physlr:
             mol_list.append((bxmol, molec_mxs))
             mol += 1
         return mol_list
+ 
+    @staticmethod
+    def split_minimizers_bx2(bx, g, bxtomxs):
+        "Partition the minimizers of the given barcode"
+        bx_match = re.compile(r'^(\S+)_\d+_\d+$')
+        mxs = bxtomxs[bx]
+        mol = 0
+        mol2 = 0
+        mol_list = []
+        while g.has_node(bx + "_" + str(mol)+ "_" + str(mol2)):
+            while g.has_node(bx + "_" + str(mol)+ "_" + str(mol2)):
+                bxmol = bx + "_" + str(mol)+ "_" + str(mol2)
+                neighbour_mxs_list = [bxtomxs[re.search(bx_match, v).group(1)] \
+                                      for v in g.neighbors(bxmol) \
+                                      if re.search(bx_match, v).group(1) in bxtomxs]
+                if not neighbour_mxs_list:
+                    neighbour_mxs_list = [set()]
+                neighbour_mxs_set = set.union(*neighbour_mxs_list)
+                molec_mxs = set.intersection(mxs, neighbour_mxs_set)
+                mol_list.append((bxmol, molec_mxs))
+                mol2 += 1
+                if not g.has_node(bx + "_" + str(mol)+ "_" + str(mol2)):
+                    for i in range(1,11):
+                        if g.has_node(bx + "_" + str(mol)+ "_" + str(mol2+i)):
+                            mol2 += i
+                            break
+            mol2 = 0
+            mol += 1
+            if not g.has_node(bx + "_" + str(mol)+ "_" + str(mol2)):
+                for i in range(1,11):
+                    if g.has_node(bx + "_" + str(mol+i)+ "_" + str(mol2)):
+                        mol += i
+                        break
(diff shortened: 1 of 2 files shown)


physlr/physlr.py
```diff
@@ -1430,21 +1430,36 @@ def physlr_count_molecules(self):
...
+                                      for v in g.neighbors(bxmol) \
+                                      if re.search(bx_match, v).group(1) in bxtomxs]
+                if not neighbour_mxs_list:
+                    neighbour_mxs_list = [set()]
+                neighbour_mxs_set = set.union(*neighbour_mxs_list)
+                molec_mxs = set.intersection(mxs, neighbour_mxs_set)
+                mol_list.append((bxmol, molec_mxs))
+                mol2 += 1
+                if not g.has_node(bx + "_" + str(mol)+ "_" + str(mol2)):
+                    for i in range(1,11):
```
```suggestion
                    for i in range(1, 11):
```
pylint is getting mad because of this whitespace

physlr/physlr.py
```diff
@@ -1430,21 +1430,36 @@ def physlr_count_molecules(self):
...
+                            mol2 += i
+                            break
+            mol2 = 0
             mol += 1
+            if not g.has_node(bx + "_" + str(mol)+ "_" + str(mol2)):
+                for i in range(1,11):
+                    if g.has_node(bx + "_" + str(mol+i)+ "_" + str(mol2)):
+                        mol += i
+                        break
+            #mol += 1
```
```suggestion
```

physlr/physlr.py
lcoombe: Also, why is this range hard-coded? Is it possible that there will be a `A_0_0` and `A_0_2` but not `A_0_1` for example?
If so, a more general solution could be to count the number of molecules per barcode (or molecules per molecule-split barcode) -- we have the function `count_molecules_per_bx`, and use that for the logic of the while loops.

jowong4: Yes to your first question. I'll have to look into that function. I hard-coded because I couldn't think of a simple way to deal with the problem and I figure 10 seems large enough to get it to work (which it did)

physlr/physlr.py
```diff
@@ -1462,19 +1478,23 @@ def physlr_split_minimizers(self):
             sys.exit(msg)
         g = self.read_graph([self.args.FILES[0]])
         bxtomxs = self.read_minimizers([self.args.FILES[1]])
+        bxtocount = self.count_molecules_per_bx(bxtomxs)
```
lcoombe: Hmm I'm not sure this works exactly as we want it to... For `count_molecules_per_bx`, it expects a dictionary of molecule -> minimizers, but you are giving it bx -> minimizers. So I think it would end up giving you an empty Counter?

jowong4: fixed

physlr/physlr.py
```diff
@@ -1467,19 +1483,24 @@ def physlr_split_minimizers(self):
             sys.exit(msg)
         g = self.read_graph([self.args.FILES[0]])
         bxtomxs = self.read_minimizers([self.args.FILES[1]])
+        molto0 = {mol : 0 for mol in g.nodes()}
+        bxtocount = self.count_molecules_per_bx(molto0)
```
lcoombe: I totally understand just doing this transformation to try to use the existing function, but I think it is really cleaner to just make a new function that does exactly what you want to do instead. (ie. iterate over the nodes in the graph, and count the number of molecules)

Who knows, having that function might actually be helpful for us in the future, and it isn't too many lines of code.

jowong4: done

jowong4:
Human
|Assembly|NG50|NGA50|# misassemblies|Genome fraction (%)|
|--|--|--|--|--|
|Original|58,105,812|10,216,403|524|86.840|
|Split minimizer based on graph + orient with more pieces|58,105,812|11,827,259|642|86.846|
|Original + arcs|58,505,102|15,922,618|424|86.840|
|Split minimizer based on graph + orient with more pieces + arcs|58,505,102|16,023,878|610|86.846|

PR #132, opened 2019-12-19
variable `python_executable`

afshinfard:
a variable to make it less painful calling our own installation of python


jowong4:
see #125 . `python_executable` or `python_interpreter`

afshinfard:
yep right, `python executable`! 
I just changed it to so it matches your changes. If you're about to push that PR I can close this, but if that may take a few days then we can push this today.

PR #139, opened 2020-01-09
physlr.py: remove unused code

jowong4:
Remove code that has been converted to c++.


lcoombe:
Looks like you need to change the test -- they still refer to some of the converted python code

PR #141, opened 2020-01-16
physlr-make: update rules to reflect new features

jowong4:
This PR contains updates that I missed when I updated the physlr-make file in the previous PRs #131 #140 . This PR also contains change to physlr-split-minimizers to work with `ext.mol.tsv` files.


lcoombe:
Can you rebase this on master??

jowong4:
Yeah I forgot to update this branch before pushing

bin/physlr-make
```diff
@@ -1021,6 +1021,10 @@ min_path_size=200
 %.map.$(ref).n10.paf.gz: %.path $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.tsv $(name)/$(ref).k$k-w$w.physlr.tsv
 	$(time) $(python) $(bin)/physlr map-paf -V$V -n10 $^ | $(gzip) >$@
 
+# Map the reference to the backbone graph generated with split minimizers and output PAF.
+%.map-split.$(ref).n10.paf.gz: %.path $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.tsv $(name)/$(ref).k$k-w$w.physlr.tsv
```
lcoombe: Where is the split minimizer input here?

bin/physlr-make
```diff
@@ -1021,6 +1021,10 @@ min_path_size=200
 %.map.$(ref).n10.paf.gz: %.path $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.tsv $(name)/$(ref).k$k-w$w.physlr.tsv
 	$(time) $(python) $(bin)/physlr map-paf -V$V -n10 $^ | $(gzip) >$@
 
+# Map the reference to the backbone graph generated with split minimizers and output PAF.
```
```suggestion
# Map the reference to the backbone graph with split minimizers and output PAF.
```
Is this correct? We're mapping with split minimizers, but the backbone graph isn't generated with split minimizers?

jowong4: Yeah your change is correct.

bin/physlr-make
jowong4: Modified it

bin/physlr-make
```diff
@@ -1021,6 +1021,10 @@ min_path_size=200
 %.map.$(ref).n10.paf.gz: %.path $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.tsv $(name)/$(ref).k$k-w$w.physlr.tsv
 	$(time) $(python) $(bin)/physlr map-paf -V$V -n10 $^ | $(gzip) >$@
 
+# Map the reference to the backbone graph with split minimizers and output PAF.
+%.map-split.$(ref).n10.paf.gz: %.path $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.overlap.m$m.mol.mol2-bcs.split.tsv $(name)/$(ref).k$k-w$w.physlr.tsv
+	$(time) $(python) $(bin)/physlr map-paf -V$V -n10 $^ | $(gzip) >$@
```
lcoombe: We have a `map-split` for outputs in BED format -- we don't need an equivalent for the split minimizers?

bin/physlr-make
```diff
@@ -103,13 +103,13 @@ arcs=false
 .PHONY: f1chr4 f1chr2R f1 fishchr25 fish physical-map scaffolds
 all: f1chr4 f1chr2R f1 fishchr25 fish
 
-$(lr).physlr.physical-map.path: $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.overlap.m$m.mol.mol2-bcs.backbone.path
+$(lr).physlr.physical-map.path: $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.overlap.m$m.ext.mol.backbone.path
```
lcoombe: What is the `ext` stand for, and it it necessary?

jowong4: Is it necessary? No, but it tells us what the type of mol separation that is used which is distributed+sqcosbin

lcoombe: So what are the other options other than `ext` that are possible?

jowong4: "					bc (biconnected componenets) is the least conservative and is only suitable for datasets with low barcode multiplicity."
 "					bc+k3 (biconnected componenets + k-3 cliques) is more conservative than bc and requires more time."
 "					distributed is a modified version of bc+k3 that is faster than bc+k3 but may be more (or even less) conservative."
 "					ext (extensive) mixes distributed with a modified version of sqcos (cosine similarity of squared adjacency matrix) which makes it more conservative."

afshinfard: `bc` ->`bc.mol` (is not in the makefile anymore)
`k3`->`k3.mol`,
`distributed`->`mol`,
`distributed+sqcosbin`->`ext`
`distributed+sqcosbin++sqcosbin++sqcosbin` -> `iterative.mol` (will be added with my recent PR)

`ext` or `iterative` may become the default molecule separation strategy we use, as they produce more reliable outputs.

lcoombe: Thanks for the clarification. 
Shouldn't this then use the `mol_strategy` variable?

afshinfard: `ext` is always `distributed+sqcosbin`. So no need to use `mol_strategy`.

In case hte user want to test his/her preferred strategy to make a phycal map or assembly he can use `mol_strategy` and there is a rule for that.
More exactly, the rule is not `distributed` -> `mol`, it is `mol_strategy` -> `mol` and the default value for `mol_strategy` is `distributed`. So the user can use this rule.

lcoombe: But how it was currently for these lines, the `.path` output would always soft-link to the file with `ext`, wouldn't it?
Looks like you made a fix, Johnathan?

jowong4: The problem is the physical map rules use ext. Its been modified to use mol and the default value is now `distributed+sqcosbin`.

bin/physlr-make
```diff
@@ -103,13 +103,13 @@ arcs=false
...
 	ln -sf $< $@
 
-$(lr).physlr.physical-map.$(ref).n10.paf.gz: $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.overlap.m$m.mol.mol2-bcs.backbone.map-split.$(ref).n10.paf.gz
+$(lr).physlr.physical-map.$(ref).n10.paf.gz: $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.overlap.m$m.ext.mol.backbone.map-split.$(ref).n10.paf.gz
```
afshinfard: So, is `ext` now the main strategy in our pipeline? and is it not possible for the user to specify a strategy of interest to make a `.physlr.physical-map.` ?
(and a reminder: `ext` runs sqcosbin on all subgraphs but `mol2-bcs` runs bcs only on juntions and the results will be different.)

bin/physlr-make
afshinfard: ... unless we want to make it the default for `.mol.`. So I changed my mind and yes we can use the variable.

bin/physlr-make
afshinfard: I guess with your latest commit my concern is answered too.

bin/physlr-make
lcoombe: Awesome -- the change looks good to me!

afshinfard:
@jowong4 
At line 188, in help, the default value for `mol_strategy` is still distributed.

```
	@echo "	mol_strategy			molecule separation strategy [distributed]. Available options are bc, bc+k3, distributed, ext."
```

afshinfard:
there we have bug
when the user set `mole_strategy` to `ext` the makefile call the python code with `--separation-strategy=ext` which is not valid and previously I meant to use it to call the python code with `distributed+sqcosbin`. so we can either remove `ext `from the help of physlr and make it a hidden keyword, or we can change the code so it translates `ext` to `distributed+sqcosbin`.
so the `ext` we are introducing in help is only valid when they want to make a `ext.mol` file

jowong4:
@afshinfard Are the new changes ok with you?

afshinfard:
> @afshinfard Are the new changes ok with you?

Yes, prefect. Thanks.

afshinfard:
sorry, just found another refer to `ext`
```
@echo "	mol_strategy			molecule separation strategy [distributed+sqcosbin]. Available options are bc, bc+k3, distributed, ext."
```

PR #142, opened 2020-01-16
Refactor physlr-map

jowong4:
To prime #141 , I have refactored physlr-map to avoid code duplication.

### DIFF
--- a/bin/physlr-make
+++ b/bin/physlr-make
@@ -979,7 +979,7 @@ min_path_size=200
 
 # Map the draft assembly to the backbone graph and output BED.
 %.backbone.map-split.$(draft).n10.bed: %.backbone.path $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.overlap.m$m.mol.mol2-bcs.split.tsv $(draft).k$k-w$w.physlr.tsv
-	$(time) $(python) $(bin)/physlr map --map-type split --map-pos 10 -V$V -n10 $^ >$@
+	$(time) $(python) $(bin)/physlr map --mx-type split --map-pos 10 -V$V -n10 $^ >$@
 
 # Map the draft assembly to the backbone graph and output BED.
 %.overlap.m$m.mol.backbone.map.$(draft).n10.bed: %.overlap.m$m.mol.backbone.path %.tsv $(draft).k$k-w$w.physlr.tsv
--- a/bin/physlr-make
+++ b/bin/physlr-make
@@ -979,7 +979,7 @@ min_path_size=200
 
 # Map the draft assembly to the backbone graph and output BED.
 %.backbone.map-split.$(draft).n10.bed: %.backbone.path $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.overlap.m$m.mol.mol2-bcs.split.tsv $(draft).k$k-w$w.physlr.tsv
-	$(time) $(python) $(bin)/physlr map-split -V$V -n10 $^ >$@
+	$(time) $(python) $(bin)/physlr map --map-type split --map-pos 10 -V$V -n10 $^ >$@
 
 # Map the draft assembly to the backbone graph and output BED.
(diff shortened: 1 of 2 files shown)


bin/physlr-make
```diff
@@ -979,7 +979,7 @@ min_path_size=200
 
 # Map the draft assembly to the backbone graph and output BED.
 %.backbone.map-split.$(draft).n10.bed: %.backbone.path $(lr).k$k-w$w.n$(minimum_barcode_multiplicity)-$(maximum_barcode_multiplicity).c2-x.physlr.overlap.m$m.mol.mol2-bcs.split.tsv $(draft).k$k-w$w.physlr.tsv
-	$(time) $(python) $(bin)/physlr map-split -V$V -n10 $^ >$@
+	$(time) $(python) $(bin)/physlr map --map-type split --map-pos 10 -V$V -n10 $^ >$@
```
```suggestion
	$(time) $(python) $(bin)/physlr map --mx-type split --map-pos 10 -V$V -n10 $^ >$@
```
Just a suggestion -- thinking because it is really an option to indicate if the minimizers are split or not

physlr/physlr.py
```diff
@@ -1750,23 +1750,16 @@ def index_minimizers_in_backbones(backbones, bxtomxs):
...
-            for pos, u in enumerate(path):
+                if Physlr.args.map_type == "all":
+                    if u not in bxtomxs:
+                        u = u.rsplit("_", 1)[0]
+                    if u not in bxtomxs:
+                        u = u.rsplit("_", 1)[0]
+                elif Physlr.args.map_type == "split":
+                    pass
+                else:
+                    print("Invalid --map-type argument", file=sys.stderr)
```
lcoombe: I think this parameter check should be before the for loop and the function should exit with `sys.exit` -- Otherwise you'd see this print statement over and over for each line in the file

physlr/physlr.py
```diff
@@ -2724,8 +2670,17 @@ def parse_arguments():
             help="ARCS scaffold pairing distance estimation file.")
         argparser.add_argument(
             "--dist-type", action="store", dest="dist_type", type=str, default="avg",
+            choices=["min", "avg", "max"],
             help="ARCS scaffold pairing distance type."
-                 "Acceptable values are: min, avg, max")
+                 "Accepted values are: min, avg, or max [avg].")
+        argparser.add_argument(
+            "--map-type", action="store", dest="map_type", type=str, default="all",
+            choices=["all", "split"],
```
```suggestion
            choices=["unsplit", "split"],
```
Trying to think of alternatives to `all` -- because it's not like we are not using all the minimizers in split, just split vs unsplit

physlr/physlr.py
```diff
@@ -1907,6 +1876,15 @@ def physlr_map(self):
         Map sequences to a physical map.
         Usage: physlr map TPATHS.path TMARKERS.tsv QMARKERS.tsv... >MAP.bed
         """
+        if self.args.map_pos < 0:
+            print("--map-pos cannot be negative", file=sys.stderr)
+            print("See physlr --help for more information", file=sys.stderr)
+            sys.exit(1)
+
+        if self.args.mx_type not in ["unsplit", "split"]:
+            print("Invlaid --mx-type argument", file=sys.stderr)
```
```suggestion
            print("Invalid --mx-type argument", file=sys.stderr)
```

physlr/physlr.py
```diff
@@ -1990,6 +1938,11 @@ def physlr_map_paf(self):
         Usage: physlr map TGRAPH.path TMARKERS.tsv QMARKERS.tsv... >MAP.paf
         """
 
+        if self.args.mx_type not in ["unsplit", "split"]:
+            print("Invlaid --mx-type argument", file=sys.stderr)
```
```suggestion
            print("Invalid --mx-type argument", file=sys.stderr)
```

physlr/physlr.py
```diff
@@ -2724,8 +2677,17 @@ def parse_arguments():
...
             "--dist-type", action="store", dest="dist_type", type=str, default="avg",
+            choices=["min", "avg", "max"],
             help="ARCS scaffold pairing distance type."
-                 "Acceptable values are: min, avg, max")
+                 "Accepted values are: min, avg, or max [avg].")
+        argparser.add_argument(
+            "--mx-type", action="store", dest="mx_type", type=str, default="all",
+            choices=["unsplit", "split"],
+            help="Type of minimizers used to map sequence to backbone graph."
+                 "Accepted values are: all or split [all].")
```
```suggestion
                 "Accepted values are: unsplit or split [unsplit].")
```

PR #146, opened 2020-01-24
sqcosbin and k3bin

afshinfard:
Tested; works;
but I have not compared to other results yet! will compare it soon.
Feel free to comment.


physlr/physlr.py
```diff
@@ -1468,7 +1468,13 @@ def detect_communities_cosine_of_squared(g, node_set, squaring=True, threshold=0
         import numpy as np
         from sklearn.metrics.pairwise import cosine_similarity
 
+        if len(node_set) < Physlr.args.skip_small:
```
jowong4: This is the skip small components to avoid oversplitting that you were talking about, right?

afshinfard: yep. when the subgraph is too small, the cosine similarity metric by nature does not work well. The default value is now set to 10 which worked well (in terms of not doing oversplitting) in December when I was testing.

afshinfard: Why? the function there is file indication a subset of barcodes (vertices), this function reads in those vertices and considers only those vertices fo stat retrieval. but if there's no input file for that, it works on the whole graph..

afshinfard:
My understanding is that this is (distributed+sqcosbin) theoretically better than distributed alone, but in practice results may most of the time look almost identical. The difference is that we are addressing things earlier and with more confidence this way.

PR #147, opened 2020-01-24
retrieve  statistics for subgraphs of interest

afshinfard:
This function now accepts a file indicating the subset of barcodes we are interested in their statistics.

*This has nothing to do with the final results of physlr, so if we can push it without testing/comparing the final results.


physlr/physlr.py
```diff
@@ -1726,19 +1726,32 @@ def subgraph_stats_process(u):
...
+                int(timeit.default_timer() - t0),
+                "Computing statistics for subgraphs of interest...",
+                file=sys.stderr)
+        else:
+            gin = self.read_graph(self.args.FILES)
+            print(
+                int(timeit.default_timer() - t0),
+                "Computing statistics of all subgraphs...", file=sys.stderr)
+            nodes_of_interest = gin
+        #Physlr.filter_edges(gin, self.args.n)
```
jowong4: remove this

physlr/physlr.py
```diff
@@ -1726,19 +1726,32 @@ def subgraph_stats_process(u):
...
-        print(
-            int(timeit.default_timer() - t0),
-            "Computing statistics of the subgraphs...", file=sys.stderr)
+        "Retrieve statistics of subgraphs."
+        nodes_of_interest = []
+        if len(self.args.FILES) > 1:
+            gin = self.read_graph([self.args.FILES[0]])
+            with open(self.args.FILES[1]) as fin:
+                for line in fin:
+                    nodes_of_interest.append(line.split()[0])
```
jowong4: Can you print one line of the file so I can know the format

physlr/physlr.py
```diff
@@ -1726,19 +1726,32 @@ def subgraph_stats_process(u):
         return Physlr.subgraph_stats(Physlr.graph, u)
 
     def physlr_subgraphs_stats(self):
-        "Retrieve subgraphs' stats."
-        gin = self.read_graph(self.args.FILES)
-        Physlr.filter_edges(gin, self.args.m)
-        print(
-            int(timeit.default_timer() - t0),
-            "Computing statistics of the subgraphs...", file=sys.stderr)
+        "Retrieve statistics of subgraphs."
```
lcoombe: Could we just add a small 'usage' comment? (similar to physlr_map, for example)

physlr/physlr.py
```suggestion
                    nodes_of_interest.append(line.strip().split()[0])
```
Just to make sure we've stripped any white space

lcoombe:
```suggestion
                    nodes_of_interest.append(line.strip().split()[0])
```
Is it possible for there to just be entry on the line? I guess goes along with Johnathan's question

afshinfard: @jowong4 @lcoombe Can be in any format by with the barcode in the first column, like:
`ACGTACGTACGT-1  100`
or just the entry alone:
`ACGTACGTACGT-1`

@lcoombe What does a call to `strip()` with no arguments do?

physlr/physlr.py
afshinfard: Why? the function there is file indication a subset of barcodes (vertices), this function reads in those vertices and considers only those vertices for stat retrieval. but if there's no input file for that, it works on the whole graph..

jowong4: I meant remove the commented out line

physlr/physlr.py
afshinfard: I just added more detail. But its a side utility; it is not used in the main pipeline.

physlr/physlr.py
afshinfard: I see :)) I'll do it.

afshinfard: done

physlr/physlr.py
lcoombe: @afshinfard - It just strips the white space from the beginning and end of the string (like newlines at the end of a line)

physlr/physlr.py
lcoombe: But you could still call `physlr subgraph-stats` right? (even if it's not in the main pipeline)
Just thinking if any of us want to do it in the future it would help to know the expected arguments
```
Usage: physlr subgraph-stats <graph .tsv> [TSV file listing barcodes]
```
(or whatever format the TSV argument is)

afshinfard: Oh I see, I thought you mean at the beginning of this function. I'll just do it.

afshinfard: Aha, Done

PR #148, opened 2020-01-25
Iterative mol sep 2

aafshinfard:
Runs molecule separation iteratively without writing to file until the end -> faster iteration.
There is a default `mol_strategy` called `iterative` that runs 3 rounds being `distributed+sqcosbin`, `sqcosbin`, and then `sqcosbin` sequentially and with different settings (see `set_settings()`)

* The ram usage is at least double in rounds 2+, so this code reduces the number of threads by a factor of 8 in rounds 2+ (as most of the time is spent on non-parallelized parts) to reduce python ram consumption in multiprocessing mode.


bin/physlr-make
```diff
@@ -1117,10 +1135,16 @@ g=$(shell awk '{x += $$2} END{print x}' $(name)/$(ref).fa.fai)
...
+
+# Separate barcodes into molecules using K3-clique community detection.
+%.iterative.mol.tsv: %.tsv
+	$(time) $(python) $(bin)/physlr molecules -V$V -t$t --separation-strategy=distributed+sqcosbin++sqcosbin++sqcosbin $< >$@
+
 # Separate barcodes into molecules using K3-clique and sqcosbin community detection.
 %.ext.mol.tsv: %.tsv
 	$(time) $(python) $(bin)/physlr molecules -V$V -t$t --separation-strategy=distributed+sqcosbin $< >$@
 
+
```
jowong4:
```suggestion

```

bin/physlr-make
```diff
@@ -1117,10 +1135,16 @@ g=$(shell awk '{x += $$2} END{print x}' $(name)/$(ref).fa.fai)
 %.k3.mol.tsv: %.tsv
 	$(time) $(python) $(bin)/physlr molecules -V$V -t$t --separation-strategy=bc+k3 $< >$@
 
+
```
jowong4:
```suggestion

```

physlr/physlr.py
```diff
@@ -1425,6 +1464,8 @@ def assign_read_molecule(mxs, moltomxs, mol_counts, bx):
     @staticmethod
     def detect_communities_biconnected_components(g, node_set):
         """Separate bi-connected components. Return components."""
+        # if len(node_set) < Physlr.args.skip_small:
+        #     return [set(node_set)]
```
jowong4: Are we not skipping small sets now?

bin/physlr-make
```diff
@@ -906,44 +910,44 @@ endif
...
 # Report the junctions in a overlap/backbone graph TSV.
 %.junctions.tsv: %.tsv
-	$(time) $(python) $(bin)/physlr report-junctions-graph --prune-branches=10 --junction-depth=1 -s0 -V$V $< >$@
+	$(time) $(python) $(bin)/physlr report-junctions-graph --prune-branches=$(prune_branches) --junction-depth=1 -s0 -V$V $< >$@
 
 # Redo molecule separation on a list of junction-causing barcodes.
-%.mol2.tsv: %.tsv %.junctions.tsv
-	$(time) $(python) $(bin)/physlr molecules -V$V -t$t --separation-strategy=bc $^ >$@
+%.mol2.tsv: %.tsv
+	$(time) $(python) $(bin)/physlr molecules -V$V -t$t --separation-strategy=bc --round=2 $^ >$@
```
jowong4: These multi rounds will result in `barcode_mol` format or `barcode_mol_mol2`

physlr/physlr.py
```diff
@@ -1519,7 +1566,7 @@ def merge_communities(g, communities, node_set=0, strategy=0, cutoff=20):
         if len(communities) == 1 and (node_set == 0 or strategy != 1):
             return communities
         if strategy == 1:  # Merge by Initializing Louvain with the communities
-            return Physlr.detect_communities_louvain(g, node_set, communities)
+            return Physlr.detect_communities_louvain(g, communities, node_set)
```
jowong4: `    def detect_communities_louvain(g, node_set, init_communities=None):` Your function definition looks like this

lcoombe:
Looks like pylint still isn't happy -- Do you want to ping Johnathan and I when CI is passing so we can review?

physlr/physlr.py
```diff
@@ -1674,62 +1799,141 @@ def physlr_molecules(self):
...
+            # Physlr.args.prune_bridges = 10
+            # Physlr.args.prune_branches = 10
+            # Physlr.args.prune_junctions = 50
+            if round > 1:
+                print(
+                    int(timeit.default_timer() - t0),
+                    "Detecting junction-causing barcodes",
+                    file=sys.stderr)
+                junctions = Physlr.identify_junctions_graph(gin)
+            # round = round + 1
```
jowong4:
```suggestion

```

physlr/physlr.py
```diff
@@ -1674,62 +1799,141 @@ def physlr_molecules(self):
...
+                n = gin.nodes[u]["n"]
+                nmolecules = 1 + max(vs.values()) if vs else 0
+                if nmolecules > 1:
+                    cumul_nmolecules += (nmolecules - 1)
+                    cumul_junctions += 1
+                for i in range(nmolecules):
+                    if round > 1:
+                        gout.add_node(f"{u}-{i}", n=n)
+                    elif round == 1:
+                        gout.add_node(f"{u}_{i}", n=n)
```
jowong4: What is this if else loop for?

physlr/physlr.py
```diff
@@ -1674,62 +1799,141 @@ def physlr_molecules(self):
...
+                # set([cluster for v, cluster in a.items()])
```
jowong4:
```suggestion

```

physlr/physlr.py
```diff
@@ -1674,62 +1799,141 @@ def physlr_molecules(self):
...
+            print(
+                int(timeit.default_timer() - t0),
+                "Identified", cumul_nmolecules, "new molecules in",
+                gin.number_of_nodes(), "barcodes and Solved",
+                cumul_junctions, "barcode re-uses.",
+                # round(gout.number_of_nodes() / gin.number_of_nodes(), 2), "mean molecules per barcode",
```
jowong4:
```suggestion

```

physlr/physlr.py
```diff
@@ -1674,62 +1799,141 @@ def physlr_molecules(self):
...
+                "Removed", num_singletons, "isolated vertices.", file=sys.stderr)
+            # gin.clear()
+            # gin = gout.copy()
+            gin = gout
+            round = round + 1
+            gout = 0
+            # Physlr.graph.clear()
+            # gout.clear()
+            # old_molecules = molecules
+            # gc.collect()
```
jowong4: remove comments

physlr/physlr.py
```diff
@@ -1674,62 +1799,141 @@ def physlr_molecules(self):
...
+
+        if 2>1:
+            # No effect on results
+            # only set to True to report the number of junctions after last round of mol-sep
+            Physlr.identify_junctions_graph(gin)
```
jowong4: I dont see the point to this code. Wouldn't it always print the number of junctions for whichever round that it is in?

afshinfard: This code prints the number of junction "after" the last round. Normally we start printing number of junction "before" each round for rounds 2+. and takes some time to do so. So by disabling it we save some time for a report that is not necessary. I will just make it condition using a parameter with default value being `Fasle`.

physlr/physlr.py
afshinfard: Currently we skip small sets only when we are running `cos` and `sqcos`. I have added this code to all function as a reminder for later improvements that we need something ~similar (but not exactly the same as this) for other functions. I can also put a single-line reminder comment if we prefer it that way.

physlr/physlr.py
afshinfard: for rounds 2+ we add a symbol other than `_`, here its `-`. I wanted to discuss this with you. Is dash confusing for some datasets?

bin/physlr-make
afshinfard: will be of format `bacode_mol1-mol2` if it's appropriate for all datasets

afshinfard:
@lcoombe @jowong4 I'm ready for reviews.
Since the results of `distributed+sqcosbin++sqcosbin++sqcosbin` are not compared to the current default, I did not set it to be the default strategy, and we still run `distributed+sqcosbin` as default but since the code is deeply reformatted, we can run `distributed+sqcosbin` again with this branch to check if the results are identical to Master.

physlr/physlr.py
jowong4: My apologies, they looked the same when I was reviewing.

bin/physlr-make
jowong4: That is gonna be a problem. Split minimizers uses regex of the form `barcode-num`

physlr/physlr.py
```diff
@@ -1653,18 +1738,59 @@ def determine_molecules_process(u):
...
+                  "\n |  junction_depth ", Physlr.args.junction_depth,
+                  " |  prune_junctions ", Physlr.args.prune_junctions,
+                  "\n | Skip small:", Physlr.args.skip_small,
+                  file=sys.stderr)
+        if round_num == 2:
+            print("Settings for round 2:", file=sys.stderr)
+
+            Physlr.args.threads = math.ceil(Physlr.args.threads / 8)
+            Physlr.args.junction_depth = 10
+            Physlr.args.cost = 0.55
```
lcoombe: These are very specific settings for the different rounds -- how did we come to them and will they remain hard-coded?

physlr/physlr.py
```diff
@@ -1674,62 +1800,129 @@ def physlr_molecules(self):
...
+                Physlr.junctions = junctions
+                Physlr.args.strategy = alg_list
+                with multiprocessing.Pool(self.args.threads) as pool:
+                    molecules = dict(pool.map(
+                        self.determine_molecules_process,
+                        progress(nodes_to_process), chunksize=100))
+                Physlr.graph = None
+            print(int(timeit.default_timer() - t0), "Identified molecules", file=sys.stderr)
+            if not molecules:
+                print(int(timeit.default_timer() - t0), "Identidied no molecules", file=sys.stderr)
```
```suggestion
                print(int(timeit.default_timer() - t0), "Identified no molecules", file=sys.stderr)
```

bin/physlr-make
afshinfard: I'll revert it back to `barcode_mol_mol2`. Is that fine?

physlr/physlr.py
```diff
@@ -1800,10 +1994,15 @@ def index_minimizers_in_backbones(backbones, bxtomxs):
         for tid, path in enumerate(progress(backbones)):
             for pos, u in enumerate(path):
                 if Physlr.args.mx_type == "unsplit":
-                    if u not in bxtomxs:
+                    iter_c = 0
+                    while u not in bxtomxs:
                         u = u.rsplit("_", 1)[0]
+                        if iter_c > 20:
```
lcoombe: What is this variable `iter_c` for?

afshinfard: we keep removing `_`s from the name of barcode until we find the barcode in the dictionary. Normally, we definitely find it after a couple of iterations but I've put this to make sure if someone uses a wrong input file we do not trap in an infinite loop!

physlr/physlr.py
afshinfard: These settings are experimental and are only used for rounds 2+ of molecule separation.
`Physlr.args.threads =..` I have mentioned in the first comment of this PR. I can alternatively set to 1.
`Physlr.args.cost = 0.55` and `Physlr.args.sqcost` are experimental and we will automatically determine the threshold for cosine similarity in a later update.
`Physlr.args.junction_depth = 10`: seems to be a good value experimentally and has low dependency on input dataset.

physlr/physlr.py
```diff
@@ -1674,62 +1800,129 @@ def physlr_molecules(self):
...
                 "Separating barcodes into molecules using the following algorithm(s):\n\t",
-                self.args.strategy.replace("+", " + "),
+                alg_list_2d,
+                # self.args.strategy.replace("++"," / ").replace("+", " + "),
                 file=sys.stderr)
 
         Physlr.filter_edges(gin, self.args.m)
+        round_num = self.args.round
+        for alg_list in alg_list_2d:
+            if Physlr.args.set_settings and len(alg_list_2d) > 1:
```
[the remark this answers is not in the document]
afshinfard: We could have used `self` since both are defined in this scope. I'll just change it.
Is there any python-specific difference between using the name of the class or `self`? I am asking to learn :D and we also have some other instances like this. lines 1849 to 1851.

physlr/physlr.py
```diff
@@ -1653,18 +1738,59 @@ def determine_molecules_process(u):
...
-        if not set(alg_list).issubset(alg_white_list):
-            exit_message = "Error: physlr molecule: wrong input parameter(s) " + \
-                      "--separation-strategy: " + str(set(alg_list) - alg_white_list)
-            sys.exit(exit_message)
         junctions = []
+
+        alg_white_list = {"cc", "bc", "bcbin", "cn2", "cn3", "k3", "k3bin", "k4",
+                          "cos", "cosbin", "sqcos", "sqcosbin", "louvain", "distributed"}
+        alg_list_2d = [t.split("+") for t in self.args.strategy.split("++")]
+        if not alg_list_2d:
```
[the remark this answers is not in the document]
afshinfard: you not defining it at all? or putting no default values for it? why not?

```
        argparser.add_argument(
            "--separation-strategy", action="store", dest="strategy", default="bc+k3",
            help="strategy for barcode to molecule separation [bc+k3]. Use a combination"
                 " of bc, k3, k3bin, cos, sqcos, sqcosbin, louvain, and distributed"
                 " concatenated via plus sign (example:bc+k3+bc)")
```

physlr/physlr.py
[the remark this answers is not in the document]
lcoombe: This is an instance method (specific to this instance of the Physlr class), so it has the access to all of the instance variables (with self). That is in contrast to the static methods, which don't have direct access to data about the class/instance. So in this case, we do have access to the instance variables.

physlr/physlr.py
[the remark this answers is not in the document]
lcoombe: Well that check is _only_ if it wasn't specified, but there is a default value for the argument. If a non-acceptable value is specified then the next check picks that up.
So I just couldn't see how that could happen?

physlr/physlr.py
lcoombe: Aren't we just changing the molecule IDs to "barcode_num"? I know we made that assumption in Johnathan's split-minimizers C++ code

physlr/physlr.py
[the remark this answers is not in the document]
afshinfard: yep, right. I'll change this and some other occurences..

physlr/physlr.py
afshinfard: I double checked with Johnathan and we will postpone this PR to make some changes for the molecule id

lcoombe: Sounds good -- yeah I think it'll make things easier to keep the molecule IDs in that form regardless of the number of rounds of molecule splitting.

afshinfard:
@jowong4 @lcoombe 
I just added a new function to reformat the molecule ids from a messy form like `ACGT_0_0_1_0_2` to a single number. Iterative molecule separation runs as is and at the end it calls the new function to reformat the molecule ids.
Will know address failed checks...

afshinfard:
and now all checks have passed!

afshinfard:
when I use cosine-similairy, sometimes I experience an issue with the python multiprocesssing:
```
/projects/btl/aafshinfard/virtuEnv/pypy3/lib-python/3/multiprocessing/semaphore_tracker.py:129: UserWarning: semaphore_tracker: There appear to be 1 leaked semaphores to clean up at shutdown
  len(cache))
```
Do you know if it may affect the results? like not processing some of the vertices?

afshinfard:
In a single test-run, I found an unexpected error:
```
8289 Identified molecules
Traceback (most recent call last):
  File "/projects/btl_scratch/aafshinfard/projects/physlr/iterative_mol_sep_2/physlr/bin/physlr", line 19, in <module>
    physlr.physlr.main()
  File "/projects/btl_scratch/aafshinfard/projects/physlr/iterative_mol_sep_2/physlr/physlr/physlr.py", line 3043, in main
    Physlr().main()
  File "/projects/btl_scratch/aafshinfard/projects/physlr/iterative_mol_sep_2/physlr/physlr/physlr.py", line 3039, in main
    getattr(Physlr, method_name)(self)
  File "/projects/btl_scratch/aafshinfard/projects/physlr/iterative_mol_sep_2/physlr/physlr/physlr.py", line 1914, in physlr_molecules
    m = gin.nodes[u]["m"]
KeyError: 'm'
make: *** [hg004.k40-w32.n100-5000.c2-100.physlr.overlap.m92.5.iterative.mol.tsv] Error 1
make: *** Deleting file `hg004.k40-w32.n100-5000.c2-100.physlr.overlap.m92.5.iterative.mol.tsv'

```
This happened in the 3rd round of molecule separation, that means the algorithm has passed through the first and second rounds successfully without this error! and in each round I do add feature `m` to any vertex I add to the graph!

PR #149, opened 2020-02-07
Physlr map c++

jowong4:
physlr map takes a long time mostly because loading in a large dictionary in python takes a long time.


src/physlr-map.cc
```diff
@@ -0,0 +1,530 @@
...
+						auto prevPair = std::make_pair(targetId, targetPos - i);
+						if (targetIdPosToQuerypos.find(prevPair) != targetIdPosToQuerypos.end()) {
+							prevVec.emplace_back(targetIdPosToQuerypos[prevPair][0]);
+						}
+					}
+
+					if (prevVec.empty()) {
+						prevVec.emplace_back(max);
+					}
+					prev = lowerMedian(prevVec);
```
lcoombe: Don't you do a check if `prevVec` is empty in `lowerMedian`? Do you still need to add `max` to the `prevVec` then?

jowong4: This is for determineOrientation. I can either do what I did up there or do this 
`if (prevVec.empty()) {
determineOrientation(max, curr, next);
}` and 2 more for the other possibilities.

There is more code duplication for the latter

lcoombe: Could you just return `max` from `lowerMedian`/`upperMedian` if prevVec is empty?

jowong4: Hmm yes, but median function giving me a max value when the input array is empty doesn't feel quite right.

jowong4:
| | Time (hh:mm:ss) | Ram (GB) |
|--|--|--|
|Python 16 threads unsplit|1:00:36|179|
|C++ 16 threads unsplit|0:10:45|58|
|Python 16 threads split|1:41:28|216|
|C++ 16 threads split|0::18:02|69|

src/physlr-map.cc
```diff
@@ -0,0 +1,561 @@
...
+uint64_t
+lowerMedian(std::vector<uint64_t> scores)
+{
+	uint64_t size = scores.size();
+
+	if (size == 0) {
+		return 0;
+	}
+
+	sort(scores.begin(), scores.end());
```
lcoombe: Just double checking because I recall us discussing for python -- it is OK that this sorts in place?

jowong4: `lowerMedian(std::vector<uint64_t> scores)` this is not passing by reference so it should be ok.

jowong4:
> I figured you were using max like an indicator variable, so it would make sense that if the vector is empty, to return that indicator/NA-like variable.
Just feels weird to check for an empty vector twice and add a random variable if it's empty instead of dealing with it in a function call?
Fine to leave at it is though if you want -- I don't feel strongly.

As I mentioned, this approach has the least lines of code so I chose it.

src/physlr-map.cc
jowong4: I'm only checking the vector once.

PR #150, opened 2020-02-12
Gantt chart

afshinfard:


afshinfard:
an example run to test it:
`Rscript -e 'rmarkdown::render("/projects/btl_scratch/aafshinfard/projects/physlr/port_code/sqcos/physlr/data/profile-physlr.Rmd", "html_document", "/projects/btl_scratch/aafshinfard/projects/physlr/port_code/sqcos/physlr/data/test.prof.html", knit_root_dir="/projects/btl_scratch/aafshinfard/projects/physlr/port_code/sqcos/physlr/data", params=list(path="/projects/btl_scratch/aafshinfard/projects/physlr/port_code/sqcos/physlr/data/"))'`

you can change the path to any path you like to test it with other files.

jowong4:
Can you add a makefile command to generate the Gantt chart?

afshinfard:
Sure @jowong4, I'll just do it.

afshinfard:
Pufff I was struggling to make the `makefile` command work! and the problem was the formatting of spaces and newlines!

Now it works,
try ` make {NAME}.prof.html path_profile={PATH}` and use a name prefix you like and determine the path for `*.time` files. by defaults it is set to `data` path for physlr

jowong4:
Hmm, are the ram values correct? It doesn't quite look correct.

jowong4:
I would suggest making the current directory the default directory. You can probably do that by assining pwd to a variable and using that.

afshinfard:
@jowong4 The ram values I extract from the `.time` file, so in case of python with multiprocessing it's the ram usage for only a single process, I can simply fix that by multiplying the ram usage with number of threads if the language is python, but I think it's better not to scale it up so we can see other bars easily.
also note that ram usage is converted into GB and also rounded (ceiling).
So is there still any other problem you see with the reports?

data/profile-physlr.Rmd
```diff
@@ -0,0 +1,384 @@
+---
+title: "Physlr profiling - time and memory"
+author: "Amirhossein Afshinfard"
+date: "2/11/2020"
```
lcoombe:
```suggestion
```

data/profile-physlr.Rmd
```diff
@@ -0,0 +1,384 @@
...
+cat(ldir)
+
+finf <- file.info(dir(path = ldir, full.names = TRUE), extra_cols = FALSE)
+sorted_finf <- finf[with(finf, order(as.POSIXct(ctime))), ]
+sorted_files <- rownames(sorted_finf)
+sorted_time_files <- sorted_files[substr(sorted_files,
+                                         nchar(sorted_files)-4,nchar(sorted_files)) == ".time"]
+
+# Or alternatively:
+#file.info("/projects/btl/jowong/github/physlr/data_hg004_stlfr/v2release/hg004_k40.histogram.time")
```
lcoombe: Was this an alternate approach you tried and decided against? We just don't want to have commented-out code

jowong4:
Just change the green like we suggested in the meeting. Thanks!

data/profile-physlr.Rmd
```diff
@@ -0,0 +1,384 @@
...
+  len = nchar(time_line)
+  colon_marks = str_match_all(time, regex(':',
+                                          ignore.case = TRUE, perl = TRUE))[[1]]
+
+  if (dim(colon_marks)[1] == 0)
+  {
+    cat(paste("Unexpected format"))
+  } else if (dim(colon_marks)[1] == 2)
+  {
+    df[i, "time_sec"] = strtoi(substr(time, nchar(time) - 1, nchar(time)), base =
```
lcoombe: Couldn't you both check which format the time string conforms to and extract the numbers in one go?

```
> n
[1] "10:40:22"
> str_match(n, regex('([:digit:]+)\\:([:digit:]+)\\:([:digit:]+)'))
     [,1]       [,2] [,3] [,4]
[1,] "10:40:22" "10" "40" "22"
```

data/profile-physlr.Rmd
```diff
@@ -0,0 +1,384 @@
...
+
+    if (length(to_be_removed2) > 1) {
+      to_be_removed2 = to_be_removed2[2:length(to_be_removed2)]
+      cat("Following file(s) will be excluded due to having NA values")
+      print(print(sorted_time_files[to_be_removed2]))
+      print(df[to_be_removed2,])
+      df = df[!to_be_removed2,]
+    }
+
+    if (dim(df)[1] == 0) {
```
lcoombe: Could we just check this at the end? Seems like it is repeated a lot

afshinfard: I keep changing the `df` dataframe, so these if-statements are checking the file for different issues. I believe that some cases would not happen, but I just wanted to take into account any possibility...

lcoombe: Hmm..But if it that statement was true at any point in this code, wouldn't it still be true in the end? And you seem to have the same error message for each of them?

afshinfard: Umm, yes I can change the message to be more specific about the problem but that's not the main point... The main reason to check it multiple times is that if the statement is true the first time, then the operations I accomplish in the `else` part are not valid and running them results in error... I mean I am avoiding run-time errors this way...

data/profile-physlr.Rmd
afshinfard: the two formats are different... How do I detect the format and extract the numbers in one go?
format 1: `HH:MM:SS`
format 2: `MM:SS.MS` which has only one `:` and also a `.`

lcoombe: You just need two regex expressions with the capture groups - that will tell you which format the string conforms to. After you do the `str_match`, you can check each for being non-NA with if-statements. Then, once you've entered the appropriate condition, you have easy access to the numbers you need.
Roughly what I was thinking:
```
hms <- str_match(...)
msms <- str_match(...)
if (is.na(hms[1,1)) {
...Parse as h,m,s..
} else if (is.na(msms[1,1]) {
...Parse as mmssms...
}
```

data/profile-physlr.Rmd
lcoombe: My point wasn't about the error messages needing to be different (I don't think that is necessary), but about the repeated code which isn't great practice. You know best all the conditions you are checking for and how realistic the situations are (I find this part of the code a bit hard to follow), so I suggest looking at possibilities for refactoring to make it more concise.

lcoombe:
@afshinfard - If you had different physlr runs in the same directory, would this script be able to distinguish the runs for the Gantt chart?

afshinfard:
@lcoombe In that situation, the code orders all the `*.time` files and if there are multiple files for a single step it will keep the one which is older. so if you have two different runs in a single directory it will report the first run. One simple way to profile different runs is to `mv` the files you want to a subdirectory and run this on that subdirectory.

data/profile-physlr.Rmd
afshinfard: I though you mean with a single call to `str_match()`. I just changed it to something like this.

data/profile-physlr.Rmd
afshinfard: Sorry I was not good at answering this...
I just reformatted the code and now it checks once for all issues and then tries to remove rows with `NA` values.

PR #151, opened 2020-02-19
connected components mol_strategy

afshinfard:
For certain stLFR datasets, even bi-connected components is too strict. I have added connected components which is a more lenient strategy.


jowong4:
Thanks!

PR #152, opened 2020-02-20
Make subgraph c++

afshinfard:
not to be merged with master.. only for code review

### DIFF
--- a/src/physlr-molecules.cc
+++ b/src/physlr-molecules.cc
@@ -321,9 +321,7 @@ make_subgraph(Graph& g, Graph& subgraph, vertexIter vBegin, vertexIter vEnd)
 {
     // //   Make a vertex-induced subgraph of graph g, based on vertices from vBegin to vEnd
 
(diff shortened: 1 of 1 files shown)


afshinfard:
I should have done this just after I did port the code.. switching between languages is painful!

This code successfully works on tests, but seems to be slow on real datasets... any suggestions?!

afshinfard:
one other alternative is to:
1- make a new subgraph with all the vertices, 
2- and then each vertex:
extract the neighbours from the original graph, and add an edge in the `subgraph` if the neighbours exists in the subgraph.

afshinfard:
@jowong4 This is not going to be merged yet! I wanted to hear your feedback on the implementation, as this is slower than the original `create_subgraph()` function!
Although we can make it faster by using multiple threads, I was wondering if you have suggestion to make it run faster on single thread too.

jowong4:
You could make an auxillary data structure - a set of edges. The graph currently stores it in a vector so to locate an edge takes some time. You would use the set of edges to check for presence.

PR #155, opened 2020-02-22
Make subgraph c++

afshinfard:


src/physlr-molecules.cc
```diff
@@ -314,6 +326,49 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
+    // check whether there exist an edge between their corresponding vertices in the graph.
+    graph_t::vertex_iterator vIter1, vIter2, vend1, vend2;
+    for (boost::tie(vIter1, vend1) = vertices(subgraph); vIter1 != vend1; ++vIter1)
+    {
+        for (boost::tie(vIter2, vend2) = vertices(subgraph); vIter2 != vend2; ++vIter2)
+        {
+            if (vIter1 != vIter2)
+            {
+		        std::unordered_map<
+		                std::pair<std::size_t,size_t>, int, pair_hash>::const_iterator got =
```
jowong4:
```suggestion
		                std::pair<std::size_t,size_t>, int, boost::hash<std::size_t,size_t>>>::const_iterator got =
```

Use that instead and remove the struct at the beginning.

src/physlr-molecules.cc
```diff
@@ -314,6 +326,49 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
+    // Iterate over pairs of vertices:
+    // check whether there exist an edge between their corresponding vertices in the graph.
+    graph_t::vertex_iterator vIter1, vIter2, vend1, vend2;
+    for (boost::tie(vIter1, vend1) = vertices(subgraph); vIter1 != vend1; ++vIter1)
+    {
+        for (boost::tie(vIter2, vend2) = vertices(subgraph); vIter2 != vend2; ++vIter2)
+        {
+            if (vIter1 != vIter2)
+            {
+		        std::unordered_map<
```
jowong4: Might want to benchmark `tsl::robin_map` to see which is faster

src/physlr-molecules.cc
afshinfard: are you sure about the efficiency? no need to test?

src/physlr-molecules.cc
afshinfard: https://tessil.github.io/2016/08/29/benchmark-hopscotch-map.html
seem to be faster. I will benchmark for out usage too.

jowong4: we aleady have the robin hash library under vendor so you can remove those files

afshinfard:
I ran this on `hg004...m92.5.tsv` with strategy being `bc` and it took 19 mins to complete:
`Elapsed (wall clock) time (h:mm:ss or m:ss): 19:38.58`
Next is to make it parallel and also add binning to it.
Then I will merge my other branches to add to this other community detection algorithms.

src/physlr-molecules.cc
```diff
@@ -20,6 +23,7 @@
 #define GIT_REVISION "pre-autotools"
 
 static uint64_t
+
```
jowong4:
```suggestion

```

src/physlr-molecules.cc
```diff
@@ -314,35 +324,86 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
+    graph_t::vertex_iterator vIter1, vIter2, vend1, vend2;
+    for (boost::tie(vIter1, vend1) = vertices(subgraph); vIter1 != vend1; ++vIter1)
+    {
+        for (boost::tie(vIter2, vend2) = vertices(subgraph); vIter2 != vend2; ++vIter2)
+        {
+            if (vIter1 != vIter2)
+            {
+		        tsl::robin_map<
+		            std::pair<std::size_t,size_t>, int,
+		                boost::hash<std::pair<size_t,size_t>>>::const_iterator got =
```
jowong4:
```suggestion
		        auto got =
```

src/physlr-molecules.cc
```diff
@@ -314,35 +324,86 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
+		                edge_set.find(
+                            std::pair<std::size_t,size_t>(
+                                subgraph[*vIter1].indexOriginal,
+                                subgraph[*vIter2].indexOriginal));
```
jowong4: Does this code work? You don't need `std::make_pair<std::size_t,size_t>` ?

src/physlr-molecules.cc
```diff
@@ -57,6 +61,7 @@ struct edgeComponent_t
 	using kind = boost::edge_property_tag;
 } edgeComponent;
 
+// this definition assumes there is no redundant edge in the undirected graph
 using graph_t = boost::subgraph<boost::adjacency_list<
```
jowong4: btw since you are making your own subgraph. you can probaly drop `boost::subgraph`

src/physlr-molecules.cc
```diff
@@ -74,6 +79,11 @@ using vertexSet_t = std::unordered_set<vertex_t>;
 using componentToVertexSet_t = std::vector<vertexSet_t>;
 using vertexToComponent_t = std::unordered_map<vertex_t, size_t>;
 using vecVertexToComponent_t = std::vector<vertexToComponent_t>;
+using vertexToIndex_t = std::unordered_map<vertex_t, size_t>; // wanna improve this? checkout boost::bimap
```
jowong4: I would recommend uint64_t instead of size_t for everything

src/physlr-molecules.cc
afshinfard: Yes I believe, as I have tested this, and I do have output of the same size to the original code. I just checked and they have the same number of lines (the `tsv` files).

src/physlr-molecules.cc
afshinfard: So, the main graph is also of this type, what is the difference?
I can later change this to make a graph based on adjacency MATRIX so I can remove a conversion step, but now its not a bottleneck.

src/physlr-molecules.cc
afshinfard: So should I change every and each occurrence of `size_t`? Is it fine if I do that in a separate PR?

src/physlr-molecules.cc
jowong4: Well the whole point of the boost subgraph wrapping the adjacency graph is to obtain the subgraph, but since we arent using boost's way to get subgraph we dont have to include it. Additionally there is an overhead there.

src/physlr-molecules.cc
```diff
@@ -314,35 +325,79 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
+	// Iterate over all pairs of vertices in the subgraph:
+	// check whether there exist an edge between their corresponding vertices in the source graph.
+	graph_t::vertex_iterator vIter1, vIter2, vend1, vend2;
+	for (boost::tie(vIter1, vend1) = vertices(subgraph); vIter1 != vend1; ++vIter1) {
+		for (boost::tie(vIter2, vend2) = vertices(subgraph); vIter2 != vend2; ++vIter2) {
+			if (vIter1 != vIter2) {
+				//		        tsl::robin_map<
+				//		            std::pair<std::uint64_t,uint64_t>, int,
+				//		                boost::hash<std::pair<uint64_t,uint64_t>>>::const_iterator
+				auto got = edge_set.find(std::pair<std::uint64_t, uint64_t>(
```
jowong4:
```suggestion
				auto got = edge_set.find(std::make_pair(
```

bit shorter

jowong4:
> I ran this on `hg004...m92.5.tsv` with strategy being `bc` and it took 19 mins to complete:
> `Elapsed (wall clock) time (h:mm:ss or m:ss): 19:38.58`
> and it used 18 GB of ram.
> With the current code it runs in almost 26 mins and with 14 GB of ram.
> 
> Next is to make it parallel and also add binning to it.
> Then I will merge my other branches to add to this other community detection algorithms.

I dont understand. What took 19 mins to complete? Python bc or the code before your changes?

afshinfard:
With the current 'master', it takes around 26 minutes (14 GB ram).
With this branch we are preparing to push, it takes around 19 minutes (18 GB ram)

afshinfard:
I just tested the last version of this code, and the `tsv` file has the same number of lines as the one I got from `master`. Also tested on the test data and it worked.

jowong4:
> With the current 'master', it takes around 26 minutes (14 GB ram).
> With this branch we are preparing to push, it takes around 19 minutes (18 GB ram)
> and now we can multi-thread (I am doing it along with binning the subgraphs in another branch) .

Very good. Thanks for doing this!

src/physlr-molecules.cc
```diff
@@ -314,35 +325,82 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
 		{ "help", no_argument, &help, 1 },
 		{ "separation-strategy", required_argument, nullptr, 's' },
 		{ nullptr, 0, nullptr, 0 }
 	};
-	for (int c; (c = getopt_long(argc, argv, "s:v", longopts, &optindex)) != -1;) {
+#if _OPENMP
+	for (int c; (c = getopt_long(argc, argv, "s:vt:", longopts, &optindex)) != -1;) {
+#else
+	for (int c; (c = getopt_long(argc, argv, "s:v:", longopts, &optindex)) != -1;) {
+#endif
```
jowong4: `for (int c; (c = getopt_long(argc, argv, "s:vt:", longopts, &optindex)) != -1;) {` Remove all that and keep this.

src/physlr-molecules.cc
```diff
@@ -314,35 +325,82 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
 		switch (c) {
 		case 0:
 			break;
 		case 's':
 			separationStrategy.assign(optarg);
 			break;
+#if _OPENMP
```
jowong4: remove

src/physlr-molecules.cc
```diff
@@ -314,35 +325,82 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
+		case 't':
+			threads = std::stoi(optarg);
+			break;
+#endif
```
jowong4: remove

PR #156, opened 2020-02-26
Omp and binning c++

afshinfard:
adding random `binning` of the subgraphs.
adding Open MP multi-threading.


afshinfard:
to check the randomness of the `unordered_set` and to see wheher its reproducible, I ran the following code multiple times:
```
int main () {
  std::unordered_set<int> myvector;

  // set some values:
  for (int i=1; i<10; ++i) unord_set.insert(i); // 1 2 3 4 5 6 7 8 9

  // print out content:
  std::cout << "unord_set contains:";
  for (std::unordered_set<int>::iterator it=unord_set.begin(); it!=unord_set.end(); ++it)
    std::cout << ' ' << *it;
```

afshinfard:
reduces the running time from 19 mins to 8 mins using 5 threads (dataset:`hg004...m92.5.tsv`). no much gain when using more threads with strategy=`bc`. But I believe when we add other strategies we will gain speed using more threads.

src/physlr-molecules.cc
```diff
@@ -323,6 +329,78 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
+	// //   randomly split the component (set of vertices) into smaller even bins
+
+	std::vector<uint64_t> components_size;
+	uint64_t neighborhood_size;
+	uint64_t components_count;
+	for (uint64_t i = 0; i < source.size(); i++) { // NOLINT
+		neighborhood_size = source[i].size();
+		components_count = ((neighborhood_size - 1) / bin_size) + 1;
+		components_size.push_back(components_count);
+	}
```
jowong4: You can use the range-based for loop here

src/physlr-molecules.cc
```diff
@@ -323,6 +329,78 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
+		auto elementIt = source[i].begin();
+		while (elementIt != source[i].end()) {
+			uint64_t length = base_com_size + yet_leftover;
+			if (--leftover == 0) {
+				yet_leftover = 0;
+			}
+
+			for (uint64_t j = 0; j < length; j++) {
+				if (counter_new >= binned_neighbours.size()) {
+					std::cerr << " WAS NOT EXPECTED 1!" << std::endl;
```
jowong4: Are you going to change these error messages to something more descriptive?

src/physlr-molecules.cc
```diff
@@ -450,18 +529,66 @@ main(int argc, char* argv[])
...
-		vertexToComponent_t vertexToComponent;
 
-		// Find neighbour of vertex and generate neighbour induced subgraph
-		auto neighbours = boost::adjacent_vertices(*vertexIt, g);
+	bool openmp = false;
+#if _OPENMP
+	openmp = true;
+#endif
+
+	if (threads > 1 && openmp) {
```
jowong4:
```suggestion
	if (threads > 1) {
```
Why do you need all that extra stuff?

src/physlr-molecules.cc
```diff
@@ -482,4 +609,4 @@ main(int argc, char* argv[])
 		std::cerr << "in sec: " << omp_get_wtime() - sTime << std::endl;
 #endif
 	}
-}
+}
```
jowong4:
```suggestion
}

```
Add an empty line here

src/physlr-molecules.cc
```diff
@@ -323,6 +329,78 @@ biconnectedComponents(graph_t& subgraph, vertexToComponent_t& vertexToComponent)
...
+	binned_neighbours.resize(new_size);
+	uint64_t counter_new = 0;
+	uint64_t base_com_size;
+	uint64_t leftover;
+
+	for (uint64_t i = 0; i < source.size(); i++) { // NOLINT
+		// Using unordered_set, we make use of its random nature and we do not shuffle randomly
+		base_com_size = source[i].size() / components_size[i];
+		leftover = source[i].size() % components_size[i];
+		uint64_t yet_leftover = (leftover ? 1 : 0);
```
jowong4:
```suggestion
		auto yet_leftover = (leftover ? 1 : 0);
```

src/physlr-molecules.cc
```diff
@@ -450,18 +529,66 @@ main(int argc, char* argv[])
...
 
-		// Find neighbour of vertex and generate neighbour induced subgraph
-		auto neighbours = boost::adjacent_vertices(*vertexIt, g);
+	bool openmp = false;
+#if _OPENMP
+	openmp = true;
+#endif
+
+	if (threads > 1 && openmp) {
+		const uint64_t array_size = boost::num_vertices(g);
```
jowong4:
```suggestion
		const auto array_size = boost::num_vertices(g);
```

src/physlr-molecules.cc
afshinfard: Does that loop access the elements in the same order this loop does? as I am pushing some values for each element into `components_size` and I later use them together so I need them to be consistent. I previously experienced inconsistency using the range-based for loop (when I tried to simulate the `i` counter for the other loop which I asked about yesterday)...

src/physlr-molecules.cc
afshinfard: mmm, I can even remove these comments as this situation is not expected at all. It was for testing when developing. what do you think?

src/physlr-molecules.cc
jowong4: Oh I see. Stick to this then. Might want to add a comment saying specifically you need it because of order in case someone in the future decides to change the code here.

src/physlr-molecules.cc
jowong4: I mean if you want to do error checking Im all for it, but if you don't think they will ever occur then you should remove it. If you do keep them, I'd prefer more descriptive error messages.

src/physlr-molecules.cc
afshinfard: You mean the `if-else` statement?
so when using multiple threads we need to make an auxiliary data structure that needs time and memory. When open mp is disabled or we are using only one thread we can skip using that data-structure and do the job faster (compared to 1-thread and using the aux data structure) and with ~ 30% less ram usage.

jowong4: Right, but why do you need the `openmp` variable.

afshinfard: Mmm, what if they compile without `openmp` but they set `-t10`?
it'd be better to output a warning message instead of this. hum?

jowong4: you can do 
```

#if ! _OPENMP
if (threads > 1) {
threads = 1;
std::cerr << ""Setting threads to 1" <<std::endl
}
#endif
 
--

Comment (jowong4) [reply src/physlr-molecules.cc]: I dont think they can even compile without openMP btw. We have that compilation flag on in the make file. It's also a requirement on the README.

File: src/physlr-molecules.cc [hunk 7]
```diff
@@ -609,4 +603,4 @@ main(int argc, char* argv[])
 		std::cerr << "in sec: " << omp_get_wtime() - sTime << std::endl;
 #endif
 	}
-}
+}
```
jowong4: you removed the newline character

PR #161, opened 2020-04-01
Physlr subgraph enable d

afshinfard:
using `bfs`, include neighbours to a certain depth in the subgraph(s)

### DIFF
--- a/bin/physlr-make
+++ b/bin/physlr-make
@@ -1139,7 +1139,7 @@ g=$(shell awk '{x += $$2} END{print x}' $(name)/$(ref).fa.fai)
 
 # Make multiple vertex-induced subgraphs
 %.subgraphs: %.tsv
-	$(python) $(bin)/physlr subgraphs -v$v -d$d -V$V -o $*.subgraphs $<
+	$(python) $(bin)/physlr subgraphs -v$v -d$d -V$V -o $*.subgraphs.$d $<
 
 # Extract statistics of the subgraphs
 %.subgraphs-stats.tsv: %.tsv
--- a/bin/physlr-make
+++ b/bin/physlr-make
@@ -1134,7 +1134,7 @@ g=$(shell awk '{x += $$2} END{print x}' $(name)/$(ref).fa.fai)
 	$(time) $(python) $(bin)/physlr molecules -V$V -t$t --separation-strategy=bc+louvain $< >$@
 
 # Make a vertex-induced subgraph
-%.subgraph.tsv: %.tsv
+%.subgraph-d$d.tsv: %.tsv
 	$(python) $(bin)/physlr subgraph -v$v -d$d -V$V $< >$@
 
 # Make multiple vertex-induced subgraphs
--- a/bin/physlr-make
+++ b/bin/physlr-make
@@ -1138,7 +1138,7 @@ g=$(shell awk '{x += $$2} END{print x}' $(name)/$(ref).fa.fai)
 	$(python) $(bin)/physlr subgraph -v$v -d$d -V$V $< >$@
 
 # Make multiple vertex-induced subgraphs
-%.subgraphs: %.tsv
+%.subgraphs.$d: %.tsv
 	$(python) $(bin)/physlr subgraphs -v$v -d$d -V$V -o $*.subgraphs.$d $<
 
 # Extract statistics of the subgraphs
--- a/bin/physlr-make
+++ b/bin/physlr-make
@@ -11,6 +11,9 @@ w=32
 # Number of threads.
 t=16
 
+# distance (depth) for extracting subgraphs
+d=1
+
 # Size of Bloom filter
 bloom_filter_size=10000000000 #10GB
 
--- a/bin/physlr-make
+++ b/bin/physlr-make
@@ -1139,7 +1139,7 @@ g=$(shell awk '{x += $$2} END{print x}' $(name)/$(ref).fa.fai)
 
 # Make multiple vertex-induced subgraphs in %.subgraphs-d$d subdirectory
 %.subgraphs-d$d: %.tsv
-	$(python) $(bin)/physlr subgraphs -v$v -d$d -V$V -o $*.subgraphs.$d $<
+	$(python) $(bin)/physlr subgraphs -v$v -d$d -V$V -o $*.subgraphs-d$d $<
 
 # Extract statistics of the subgraphs
 %.subgraphs-stats.tsv: %.tsv
--- a/bin/physlr-make
+++ b/bin/physlr-make
@@ -1137,8 +1137,8 @@ g=$(shell awk '{x += $$2} END{print x}' $(name)/$(ref).fa.fai)
 %.subgraph-d$d.tsv: %.tsv
 	$(python) $(bin)/physlr subgraph -v$v -d$d -V$V $< >$@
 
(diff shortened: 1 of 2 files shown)


physlr/physlr.py
```diff
@@ -967,15 +967,19 @@ def physlr_flesh_backbone(self):
 
     def physlr_subgraph(self):
         "Extract a vertex-induced subgraph."
-        if self.args.d not in (0, 1):
-            sys.exit("physlr subgraph: error: Only -d0 and -d1 are currently supported.")
+        if self.args.d > 50:
```
lcoombe, line 970: 50 seems to be a 'magic number' here?
It would be better for it to be a variable somewhere with default 50 I think -- easier to change in the future

physlr/physlr.py
```diff
@@ -998,10 +1002,26 @@ def physlr_subgraphs(self):
             os.makedirs(self.args.output)
         for u in progress(vertices):
             vertices_u = set()
+            u_mol = set()
+            if u in g:
+                u_mol.add(u)
+            else:
+                for i in range(100):
```
lcoombe, line 1009: 100 is another 'magic number'? Is this a hack to catch all the molecules?
I know @jowong4  and I have discussed other approaches to avoid doing this in the past with his code

physlr/physlr.py
afshinfard: Actually 50 is a fixed limit I defined for `self.args.d` as larger values make it slow, and I don't see a usage for it for now; but I can remove this limitation. This is not an important parameter to be a variable...

physlr/physlr.py
afshinfard: Yes I am trying to include all the molecules; I can alternatively continue this loop infinitely until visiting molecule which is not present in the graph.. hum? 
@jowong4 Do you remember the solution you used for the same issue before?

physlr/physlr.py
lcoombe: Having a 'magic number' isn't good coding practice. 
If we keep these references to the number, it was a number you decided on, but that doesn't mean we won't change it in the future, and I didn't understand why it was that particular value reading the code.

physlr/physlr.py
lcoombe: If I recall correctly (Correct me if I'm wrong Johnathan), we had an initial step to count the number of molecules per barcode, so you could stop once you'd seen all the molecule of a barcode

jowong4: Well yes, that was the idea but it doesn't actually work, because `xxx_1_0` doesn't have to exist for `xxx_1_1` to exist. The c++ code doesn't have this problem but I'll have to look into it again.

lcoombe: So we never fixed that issue with the double molecule IDs? (ie. re-numbering if needed so avoid the double molecule nums) Sorry - my memory is fuzzy

jowong4: That said, if you can guarantee there is only one underscore, then the method would work, which I believe is the case here.

jowong4: > ble molecule IDs? (ie. re-numbering if needed so avoid the double molecule nums

not yet

afshinfard: So if I continue the loop until it sees a barcode that is not present in the graph (assuming a single underscore), that does the same job. Its like I am counting the number of molecules but by checking them in the graph.

physlr/physlr.py
afshinfard: Umm, you are totally right about not putting `magic numbers`, but the solution would be to totally remove this restriction on `self.args.d`... because the algorithm can theoretically to the job for any value of `self.args.d`... I mean there is no need for any if-statements on the parameter `self.args.d`. I am not against making that a variable, but now I think a fixed value or even a variable does not make that much sense here... the only problem is that it may be slow... but it will do the job... and I have already tested a slow case where it returned the whole graph as the subgraph, and it was fine.

physlr/physlr.py
afshinfard:
```
        for u in progress(vertices):
            vertices_u = set()
            u_mol = set()
            if u in g:
                u_mol.add(u)
            else:
                mol = 0
                to_mol = u + "_" + str(mol)
                while to_mol in g:
                    u_mol.add(to_mol)
                    mol = mol + 1
                    to_mol = u + "_" + str(mol)```

File: physlr/physlr.py (hunk 1 above)
Comment (jowong4) [reply physlr/physlr.py:970]: I think what Lauren meant is you should make it a parameter.

Comment (afshinfard) [reply physlr/physlr.py]: I just added a parameter called 'self.args.d_max'

File: physlr/physlr.py (hunk 2 above)
Comment (afshinfard) [reply physlr/physlr.py]: changes are made.

File: bin/physlr-make [hunk 3]
```diff
@@ -1134,12 +1134,12 @@ g=$(shell awk '{x += $$2} END{print x}' $(name)/$(ref).fa.fai)
 	$(time) $(python) $(bin)/physlr molecules -V$V -t$t --separation-strategy=bc+louvain $< >$@
 
 # Make a vertex-induced subgraph
-%.subgraph.tsv: %.tsv
+%.subgraph-d$d.tsv: %.tsv
```
lcoombe, line 1137: Could we add a default setting for `d` at the top of the Makefile, with the other variable declarations? Just so we can have a comment there as well reminding our future selves what this parameter is.

physlr/physlr.py
```diff
@@ -998,10 +1008,26 @@ def physlr_subgraphs(self):
             os.makedirs(self.args.output)
         for u in progress(vertices):
             vertices_u = set()
+            u_mol = set()
+            if u in g:
+                u_mol.add(u)
+            else:
+                mol = 0
+                to_mol = u + "_" + str(mol)
```
lcoombe, line 1016: Solution looks good to me -- just double checking that it wouldn't be possible for there to be a molecule ID skipped in the input you are using? (ie. A_1, A_3 but no A_2).

bin/physlr-make
afshinfard: Sure, will do

physlr/physlr.py
afshinfard: Um, I am not 100% sure, theoretically this case is not possible as we sort the communities by their size and then make molecules. and at the end we remove singletons. I believe singletons must be last instances. But will check the files to see if I can find such a case.

afshinfard: I just checked a `.mol.tsv` file of `10X Chromium` which has higher barcode multiplicity. checking 100000 cases I found no occurrence of a skipped ID.
```
[aafshinfard@hpce704 data]$ head /projects/btl/jowong/github/physlr/data_hg004_ntHits/bf_repetitive/threshold90/hg004.k40-w32.n100-5000.c2-x.physlr.overlap.n67.k3.mol.tsv -n100000 > ./head.tsv
[aafshinfard@hpce704 data]$ cat head.tsv | tail -n999999 > head2.tsv
[aafshinfard@hpce704 data]$ cat head2.tsv | awk '{print $1}' > head.tsv
[aafshinfard@hpce704 data]$ head head.tsv | cut -d'_' -f2 | awk '($1 > prev+1) { print prev } { prev = $1 }'
```

bin/physlr-make
afshinfard: done

PR #163, opened 2020-08-24
fix typo

afshinfard:


afshinfard:
Sure, I'll do that after I finish another task

PR #164, opened 2020-10-04
Annotate graph bxmap

afshinfard:


physlr/physlr.py
```diff
@@ -2109,6 +2109,51 @@ def chr_int(x):
             return None
         return int(x[3:]) if x.startswith("chr") else int(x)
 
+    def physlr_annotate_graph_bxtochr(self):
+        """
+        Annotate a graph with a file of barcode to chromosome mappings.
+        The mapping file includes two values per line separated by tab: barcode an chromosome.
```
lcoombe, line 2115:
```suggestion
        The mapping file includes two values per line separated by tab: barcode and chromosome.
```

physlr/physlr.py
```diff
@@ -2109,6 +2109,51 @@ def chr_int(x):
...
+        Usage: physlr physlr-annotate-graph-bxtochr GRAPH PATH MAPPINGS_FILE... >ANNOTATED-GRAPHVIZ
```
lcoombe, line 2116: Your usage command below says that only the graph and mappings file are required? So no PATH?
```suggestion
        Usage: physlr physlr-annotate-graph-bxtochr GRAPH MAPPINGS_FILE... >ANNOTATED-GRAPHVIZ
```

physlr/physlr.py
```diff
@@ -2109,6 +2109,51 @@ def chr_int(x):
...
+            sys.exit("physlr annotate-graph-bxtochr: error: two file arguments are required")
+        graph_filenames = [self.args.FILES[0]]
+        map_filename = self.args.FILES[1]
+
+        g = self.read_graph(graph_filenames)
+        Physlr.remove_small_components(g, self.args.min_component_size)
+
+        # Read barcode to chromosome data into a dictionary
+        bxtochr = {}
+        with open(map_filename) as mapfilename:
```
lcoombe, line 2129: This is nitpicky, but `map_filename` and `mapfilename` are super similar variable names - perhaps `fin` or `map_file_in` or something is better to disambiguate them more?

physlr/physlr.py
```diff
@@ -2109,6 +2109,51 @@ def chr_int(x):
...
+            for line in mapfilename.readlines():
```
lcoombe, line 2130:
```suggestion
            for line in mapfilename:
```
Using `readlines()` is less efficient since it reads the whole file contents into a list vs. iterating over the file line by line

physlr/physlr.py
```diff
@@ -2109,6 +2109,51 @@ def chr_int(x):
...
+                bx_chr = line.split()
```
lcoombe, line 2131:
```suggestion
                bx, chr = line.strip().split()
```
I always like to add a `strip` in there to get rid of any pesky white space. Also thinking you could unpack the values to the `bx` and `chr` variables, and then use them in the next line. Just nice for readability.

lcoombe (approved):
Look good! Just a few suggestions.

PR #165, opened 2020-10-10
Switch indexlr and makebf to btllib

vlad0x00:
See https://github.com/bcgsc/ntJoin/pull/40 for new indexlr changes.

The makebf change is due to indexlr using btllib Bloom filter.


src/btllib/extras/python/btllib.py
```diff
@@ -0,0 +1,89 @@
...
+    """Class decorator for adding a metaclass to a SWIG wrapped class - a slimmed down version of six.add_metaclass"""
+    def wrapper(cls):
+        return metaclass(cls.__name__, cls.__bases__, cls.__dict__.copy())
+    return wrapper
+
+
+class _SwigNonDynamicMeta(type):
+    """Meta class to enforce nondynamic attributes (no new attributes) for a class"""
+    __setattr__ = _swig_setattr_nondynamic_class_variable(type.__setattr__)
+
```
jowong4, line 62: is this meant to have so many empty lines

src/btllib/include/btllib/graph.hpp
```diff
@@ -0,0 +1,96 @@
+// #ifndef BTLLIB_GRAPH_HPP
```
jowong4, line 1: If this isn't ready (everything is commented out), shouldn't you just leave it out of the PR?

schutzekatze: As btllib is still being worked on that file exists there and is not a specific to this PR. I guess your comment is more relevant to btllib itself.

src/btllib/extras/python/btllib.py
schutzekatze: It's an automatically generated file by SWIG for btllib wrappers., so I have no control over it. It just comes with btllib, but it's not relevant to this PR.

PR #170, opened 2021-05-15
Update README.md

afshinfard:
Quick instruction on calculating assembly-like metrics for the physical map.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ cd physlr/src && make install PREFIX=/opt/physlr
 
 ## Generating Physlr Physical Map with stLFR reads
 
-To construct a physical map, you need linked reads from 10X Genomics or stLFR. In addition, to visualize the correctness and contiguity of the physical map, you need a reference genome.
+To construct a physical map, you need linked reads from 10X Genomics or stLFR. In addition, to visualize the correctness and contiguity of the physical map, you need a reference genome (optional).
 In this example, the linked reads and reference genome are called `linkedreads.fq.gz` and `reference.fa`, respectively. The linked reads are from stLFR so we specify `protocol=stlfr` to use the default value for stLFR reads.
 
 ```
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ bin/physlr-make physical-map lr=linkedreads ref=reference protocol=stlfr
 
 If you provide a reference genome, Physlr automatically maps the physical map to the input reference and outputs a PAF file. You can run the following command which uses the PAF file to calculate assembly-like quality metrics for the physical map.
 ```
-bin/physlr-make linkedreads.physlr.physical-map.reference.n10.qpos.chain.metrics.tsv lr=linkedreads ref=reference
+bin/physlr-make map-quality lr=linkedreads ref=reference
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ In this example, the linked reads dataset is called `linkedreads.fq.gz`. The lin
 cd experiment
 bin/physlr-make physical-map lr=linkedreads protocol=stlfr                # Constructs the physical map
 ```
-You also have the option to provide a reference genome (with `ref`) for Physlr to evaluate the physical map. Assuming the reference is called `reference.fa` you can run the following command for the previous example:
+You also have the option to provide a reference genome (with `ref`) for Physlr to evaluate the physical map. Assuming the reference is called `reference.fa`, you can run the following command for the previous example:
 ```
 cd experiment
 bin/physlr-make physical-map lr=linkedreads ref=reference protocol=stlfr  # Constructs the physical map and reference-based evaluations for it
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -52,6 +52,12 @@ cd experiment
 bin/physlr-make physical-map lr=linkedreads ref=reference protocol=stlfr
 ```
 
+If you provide a reference genome, Physlr automatically maps the physical map to the input reference and outputs a PAF file. You can run the following command which uses the PAF file to calculate assembly-like quality metrics for the physical map.
```
lcoombe, line 55: I'd rearrange these sentences to make it more clear that providing a reference allows physlr to calculate the quality metrics for physical map. 
Something like:

If you provide a reference genome (with `ref`), physlr automatically calculates assembly-like quality metrics for the physical map. Physlr automatically maps the physical map to the input reference, and outputs a PAF file that is parsed to calculate these metrics

README.md
```diff
@@ -52,6 +52,12 @@ cd experiment
...
+```
+bin/physlr-make linkedreads.physlr.physical-map.reference.n10.qpos.chain.metrics.tsv lr=linkedreads ref=reference
```
lcoombe, line 57: The filename for this target is dependent on the `lr` and `ref` setting right? Might be easier to make an a 'phony' target called 'ref-metrics' or something like that in the Makefile, so a user wouldn't have to manually adjust the filename?

afshinfard: Done; with the current code Physlr generates this file automatically but one can still run this (with a phony target). Please see updates.

README.md
```diff
@@ -44,14 +44,20 @@ cd physlr/src && make install PREFIX=/opt/physlr
 
 ## Generating Physlr Physical Map with stLFR reads
 
-To construct a physical map, you need linked reads from 10X Genomics or stLFR. In addition, to visualize the correctness and contiguity of the physical map, you need a reference genome.
+To construct a physical map, you need linked reads from 10X Genomics or stLFR. In addition, to visualize the correctness and contiguity of the physical map, you need a reference genome (optional).
```
jowong4, line 47: Is the "optional" needed?

afshinfard: @jowong4 I wanted to somehow remind the user that this is done _de novo_ and the reference is optional for those purposes mentioned. We can alternatively rephrase and add _de novo_ somewhere and remove this.

lcoombe: On that train of thought - I wonder if we are best to only have the first sentence here, and an example command _without_ a reference genome. Then, after that, we could say that you can visualize stats with a reference, and show an example of running in that way? 
Just thinking that we want to make it clear how to run physlr without a reference, which may be the more common use-case. Don't want someone thinking that it's reference-based.

lcoombe:
@afshinfard - I just wasn't clear looking at this PR - Is generating the map quality stats currently default when running physlr? 
If so, how would a user run `physlr` without a reference genome?

afshinfard:
@lcoombe So, the user can define a `ref` or not. It is optional. If the input the `ref`, then Physlr generates a physical-map and a metrics file and some visualizations. If they don't, then only a physical-map will be generated and they can later generate those other files using the `map-quality` phony by providing a reference.

lcoombe:
> @lcoombe So, the user can define a `ref` or not. It is optional. If the input the `ref`, then Physlr generates a physical-map and a metrics file and some visualizations. If they don't, then only a physical-map will be generated and they can later generate those other files using the `map-quality` phony by providing a reference.

Great! I think that's how it should be. I think just adding in a relevant example showing launching physlr _without_ a reference in the README is still a good idea - because I wasn't sure from just looking at the README.

afshinfard:
@lcoombe Good point, I agree, will change the example in the README to come with and without a `ref` for the `physical-map`

README.md
afshinfard: Yeah, I was just editing and I was thinking the same way you just said... I'll switch to that one in a moment

afshinfard: @lcoombe done, you can check the changes now, thank you

afshinfard:
sorry for the many small changes; it's ready to be reviewd.

README.md
```diff
@@ -44,19 +44,40 @@ cd physlr/src && make install PREFIX=/opt/physlr
...
 ## Generating Physlr Physical Map with stLFR reads
 
-To construct a physical map, you need linked reads from 10X Genomics or stLFR. In addition, to visualize the correctness and contiguity of the physical map, you need a reference genome.
-In this example, the linked reads and reference genome are called `linkedreads.fq.gz` and `reference.fa`, respectively. The linked reads are from stLFR so we specify `protocol=stlfr` to use the default value for stLFR reads.
+To construct a physical map _de novo_, you need linked reads (from 10X Genomics or MGI stLFR). 
 
+In this example, the linked reads dataset is called `linkedreads.fq.gz`. The linked reads are from stLFR so we specify `protocol=stlfr` to use the default value for stLFR reads.
+
+```
+cd experiment
```
lcoombe, line 52:
```suggestion
cd experiment # Change to working directory 
```
Just thinking to make it extra clear what you mean by this line?

lcoombe (approved):
Look good! I think it's nice and clear now :)
Just one more small suggestion from me.

README.md
afshinfard: I agree, thank you for pointing out that things were unclear

PR #174, opened 2022-06-18
Fixes

aafshinfard:


azure-pipelines.yml
```diff
@@ -3,7 +3,7 @@
 jobs:
 - job: Test
   pool:
-    vmImage: Ubuntu 16.04
+    vmImage: 'ubuntu-18.04'
```
lcoombe, line 6: What about changing to `ubuntu-latest`? Just would mean that we don't have to keep updating the image as older ones become deprecated?

lcoombe (approved):
Looks good - just a couple small comments!

azure-pipelines.yml
aafshinfard: I can check that again, but I remember I needed many changes in the code to happy `clang` for that change and I decided to go with the easy-fix for now. Also, some of the issues I could troubleshoot quicker because I had a pointer that the os version differs. If it was set to `ubuntu-latest` I would not have had that pointer.

setup.py
```diff
@@ -5,16 +5,16 @@
 
 setuptools.setup(
     name="physlr",
-    version="0.1.0",
-    author="Shaun Jackman",
-    author_email="sjackman@gmail.com",
-    description="Physical Mapping of Linked Reads",
+    version="1.0.0",
+    author="Amirhossein Afshinfard",
+    author_email="aafshinfard@gmail.com",
```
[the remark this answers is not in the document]
aafshinfard: Sure,
Should I change it later when I am graduating?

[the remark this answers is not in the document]
lcoombe: Yup, that sounds fine

azure-pipelines.yml
lcoombe: Ok yeah, if it's a problem, it's fine - just something that could become an issue in the future. 
The current version for `ubuntu-latest` is here: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml

PR #175, opened 2022-06-20
Logo

aafshinfard:
Just adding logos to the repo.
Will release v1.0.1 to fix conda afterwards


lcoombe (commented):
Shall we bump the version number everywhere? For ex. see in `physlr.py` it hasn't been updated?

aafshinfard:
> Shall we bump the version number everywhere? For ex. see in `physlr.py` it hasn't been updated?

Good catch, thanks!
I fixed it and other instances of `version`. That should be it.

lcoombe:
> > Shall we bump the version number everywhere? For ex. see in `physlr.py` it hasn't been updated?
> 
> Good catch, thanks! I fixed it and other instances of `version`. That should be it.

Great! Just checking - do we want to have that version for the C++ executables too (right now looks like they just reference a git commit)?

aafshinfard:
> Great! Just checking - do we want to have that version for the C++ executables too (right now looks like they just reference a git commit)?

Since users don't run them directly I think we don't need them.
Also, not sure what `pre-autotools` does here, but they don't print a version but just some information like:
```
$ ./physlr-overlap --version
physlr-overlap (physlr) pre-autotools
Written by Justin Chu.

Copyright 2019 Canada's Michael Smith Genome Science Centre

lcoombe:
Ok! Again not essential if you just want to push this release out, but I think we shouldn't have a `--version` option then for the C++ code if you don't want to have versions associated with them? `pre-autotools` is probably just indicating that this was prior to any sort of official release.

aafshinfard:
Yep, I agree that we need to do a lot of code cleaning. As you said I am just getting this PR through to fix `conda`. Best is if we save that code cleaning for later as Physlr is just published and people would benefit from a `conda` installation.

lcoombe (approved):

PR #181, opened 2022-07-20
fix readme info

aafshinfard:
I think we need to move this part to conda installation instructions.
Also, we ended up not including pypy3 in the dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ physlr [OPTION]... python_executable=python3
 ```
 You can install pypy3 using conda:
 ```
-conda install -c conda-forge pypy3.8 # Change 3.8 to your conda envs python version (3.6 to 3.9 are supported)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,6 +48,16 @@ Physlr can generate complmentary reports (included in the pipeline by default) -
...
 conda install -c conda-forge r-ggplot2
 ```
+Additionally, we recommend using `pypy3` over regular python3 for speed. 
+pypy v3 or `pypy3` is the default python executable for Physlr. To switch to other executables set the `python_executable` argument:
+```
+physlr [OPTION]... python_executable=python3
+```
+You can install pypy3 using conda:
+```
+conda install -c conda-forge pypy3.8 # Change 3.8 to your conda envs python version (3.6 to 3.9 are supported)
```
lcoombe, line 58:
```suggestion
conda install -c conda-forge pypy3.8 # Change specified version based on your conda environment's python version (3.6 to 3.9 are supported)
```
Just a suggestion to make the comment a bit more clear

PR #192, opened 2023-09-21
filter-edges-m

aafshinfard:
So, we no longer (since before physlr v1) have column `n` in the `.tsv` files so these commands I am omitting will return an empty file. As this functionality is still needed for developers, I am keeping one command that filters `m`, but I use parameter `n` for it as m is already allocated to filter `m%` of lowest weighted edges. may not be the vest solution but couldn't think of any other alternatives.


aafshinfard:
closing due to failing checks, will push another time.

